PR #118, opened 2017-09-13
Example tests + log vpp connect time 

lukasmacko:
vpp-agent-ctl silent log

updated cn-infra + govpp


.travis.yml
```diff
@@ -22,7 +22,7 @@ script:
   - make lint
   - make
   - make test-cover
-  - make test-examples || true
+  - sudo make test-examples
```
jozef-slezak: it is fine to keep testing of examples optional

PR #140, opened 2017-09-28
Modify vEth config

lukasmacko:
vEth config allows to define the name of the interface in the host OS different from the one used internally by th agent

### DIFF
--- a/plugins/linuxplugin/interface_config.go
+++ b/plugins/linuxplugin/interface_config.go
@@ -33,6 +33,7 @@ import (
 
 	"strings"
 
+	"github.com/ligato/cn-infra/logging"
 	"github.com/ligato/cn-infra/servicelabel"
 	"github.com/ligato/cn-infra/utils/addrs"
 	"github.com/ligato/vpp-agent/plugins/linuxplugin/ifaceidx"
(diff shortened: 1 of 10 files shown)


plugins/linuxplugin/interface_config.go
```diff
@@ -503,24 +514,24 @@ func (plugin *LinuxInterfaceConfigurator) removeObsoleteVeth(nsMgmtCtx *linuxcal
 
 // addVethInterface creates a new VETH interface with a "clean" configuration.
 func (plugin *LinuxInterfaceConfigurator) addVethInterface(nsMgmtCtx *linuxcalls.NamespaceMgmtCtx, iface *intf.LinuxInterfaces_Interface, peer *intf.LinuxInterfaces_Interface) error {
-	err := plugin.removeObsoleteVeth(nsMgmtCtx, iface.Name, iface.Namespace)
+	err := plugin.removeObsoleteVeth(nsMgmtCtx, iface.HostIfName, iface.Namespace)
```
milanlenco: removeObsoleteVeth() needs both the logical name (for calls to plugin.ifIndexes) and the physical name (for linuxcalls).

Also please remove:
plugin.ifIndexes.UnregisterName(peerName)
from removeObsoleteVeth(). It requires the peer logical name and not the physical one obtained with linuxcalls.GetVethPeerName(). There are always successive calls to removeObsoleteVeth() for both ends anyway.

PR #156, opened 2017-10-09
Set event listener for vpp

pgubka:
The pusrpose is to kill supervisord on vpp failure, which
subsequently stops the docker container.

Signed-off-by: Peter Gubka <pgubka@cisco.com>

### DIFF
--- a/docker/prod_vpp_agent/supervisord.conf
+++ b/docker/prod_vpp_agent/supervisord.conf
@@ -15,6 +15,8 @@ autorestart=true
 redirect_stderr=true
 priority=2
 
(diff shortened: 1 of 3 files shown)


docker/prod_vpp_agent/supervisord.conf
```diff
@@ -14,3 +14,7 @@ command=/bin/vpp-agent --etcdv3-config=/opt/vpp-agent/dev/etcd.conf --kafka-conf
 autorestart=true
 redirect_stderr=true
 priority=2
+
+; [eventlistener:vpp_not_running]
```
rastislavszabo: Please add a comment explaining what this does when uncommented

PR #183, opened 2017-10-23
SPOPT-1690 REST API for VPP CLI

mpundlik:


plugins/restplugin/rest_handlers.go
```diff
@@ -0,0 +1,300 @@
...
+			res, err := ifplugin.DumpInterfaces(plugin.Deps.Log, ch, nil)
+			if err != nil {
+				plugin.Deps.Log.Errorf("Error: %v", err)
+				formatter.JSON(w, http.StatusInternalServerError, nil)
+			} else {
+				plugin.Deps.Log.Debug(res)
+				formatter.JSON(w, http.StatusOK, res)
+			}
+		}
+		defer ch.Close()
```
VladoLavor: you should call defer right after the channel is created (on line 37 in this case). Otherwise, if something goes wrong in the if/else statement (nil pointer, panic, etc), defer won't be called and the channel remains open

check in every method

mpundlik: done

PR #189, opened 2017-10-25
Update build-vpp.sh script for travis

ondrej-fabry:
This should fix build issues that occur when the VPP version is updated and Travis tries to use old cached VPP deb packages from previous build, thus failing on step `make test-examples`. 
Now the script will check the VPP cache folder if it contains version that doesn't match and removes it.

I also changed `VPP_COMMIT` variable to retrieve the commit hash directly from vpp submodule. 
Now it is only required to change version at one place during VPP version change, by updating vpp submodule to desired version.

I also added `git -> depth` parameter to **5** in Travis config to save some time when cloning the repository since by default it retrieves **50**. This decreases cloning step by ~10 seconds.

### DIFF
--- a/scripts/build-vpp.sh
+++ b/scripts/build-vpp.sh
@@ -2,19 +2,20 @@
 set -e
 
 VPP_CACHE_DIR=$HOME/build-cache/vpp
+
 VPP_COMMIT=`git submodule status vpp | awk '{print $1}'`
 VPP_IMG_TAG=`echo ${VPP_COMMIT} | cut -c1-7`
 
+# check if cache folder contains same version
 if [ -d "$VPP_CACHE_DIR" ] && [ ! -f ${VPP_CACHE_DIR}/vpp_*${VPP_IMG_TAG}*.deb ]; then
-    echo "Found different VPP version in cache, removing folderr"
+    echo "Removing cached VPP folder with different version"
     rm -rf "$VPP_CACHE_DIR"
 fi
 
-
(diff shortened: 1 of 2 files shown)


scripts/build-vpp.sh
```diff
@@ -2,11 +2,18 @@
 set -e
 
 VPP_CACHE_DIR=$HOME/build-cache/vpp
-VPP_COMMIT="92b0275a364022af6ab828dfac83e38c0117cfe6"
-VPP_IMG_TAG="92b0275"
+
+VPP_COMMIT=`git submodule status vpp | awk '{print $1}'`
```
lukasmacko: On the one hand this is definitely better solution than hardcoding the commit id, on the other we reuse this script in `contiv/vpp` project where vpp-agent is in `vendor` dir, thus there is no submodule. 

What about supporting "fallback" variable (i.e. allow to define commit manualy) if there is no submodule?

ondrej-fabry: Using fallback variable only if there is no submodule would defeat the intended purpose of preventing commits where variable and submodule don't match. Then it would be better to use the hard coded variable as before to not confuse anyone.

As I was looking for the most suitable solution I could not find any real usage or dependents for the vpp submodule. Is there even any?

ondrej-fabry:
I changed the `VPP_COMMIT` variable back to the hard coded one so the rest of the changes can be merged.

PR #215, opened 2017-11-03
Add tap int crud test

DasaSimkova:


tests/robot/suites/crud/tap_crud.robot
```diff
@@ -109,3 +94,22 @@ TestSetup
 TestTeardown
     Make Datastore Snapshots    ${TEST_NAME}_test_teardown
 
+Check TAP interface State
```
msestak77: remove this Keyword from this file into the library

msestak77: also change the input instead of specific values  for checking use array of @{desired_states}  ( more flexible for reusing )

tests/robot/suites/crud/tap_crud.robot
```diff
@@ -19,88 +19,73 @@ ${ENV}=              common
 ${NAME_TAP1}=        vpp1_tap1
```
msestak77: remove "traffic" from commit message as these are only crud tests

PR #216, opened 2017-11-03
Add rx-mode interface settings for interfaces.

jgloncak:
interfaces.proto file now containes attribute rx_mode.
It is possible to set one of rx mode (from vpp vnet/interface.h
vnet_hw_interface_rx_mode) on interface.

TODO: support changing of rx mode (change on existing interface)

REMARK: this is tested with VPP v18.01-rc0~159-gad8015b~b3096, not
one which is specified in project vpp submodule


plugins/defaultplugins/ifplugin/interface_config.go
```diff
@@ -233,6 +233,12 @@ func (plugin *InterfaceConfigurator) ConfigureVPPInterface(iface *intf.Interface
 		}
 	}
 
+	// set rx mode
+	err = vppcalls.SetRxMode(ifIdx, iface.RxMode, plugin.Log, plugin.vppCh, nil  )
```
VladoLavor: From the (very brief) VPP documentation:

This command is used to assign the RX packet processing mode (polling, interrupt, adaptive) of the a given interface, and optionally a given queue. If the 'queue' is not provided, the 'mode' is applied to all queues of the interface. **Not all interfaces support all modes**. To display the current rx-mode use the command 'show interface rx-placement'.

In proto file, rx mode is not bound to any specific interface type. So you can technically set any mode to any interface. What happens if you try to set a mode to tap (which should not be supported)?  Maybe you should try which mode can be set with some interface type and reflect it in the implementation.

plugins/defaultplugins/ifplugin/model/interfaces/interfaces.proto
```diff
@@ -19,6 +28,7 @@ message Interfaces {
         bool enabled = 4;
         string phys_address = 5;
         uint32 mtu = 6;
+        RxModeType rx_mode = 7;
```
VladoLavor: add support for queues. Maybe you can turn RxModeType field to message like this:

message RxMode {
      RxModeType rx_mode
      QueueID uint32
      QueueIDValid uint32
}

plugins/defaultplugins/ifplugin/interface_config.go
```diff
@@ -173,6 +175,10 @@ func (plugin *InterfaceConfigurator) ConfigureVPPInterface(iface *intf.Interface
 		ifIdx, pending, err = plugin.afPacketConfigurator.ConfigureAfPacketInterface(iface)
 	}
 
+	var wasError error
+	//rx mode
+	wasError = plugin.configRxModeForInterface(iface, ifIdx)
```
VladoLavor: call it after pending (line 185)

plugins/defaultplugins/ifplugin/interface_config.go
```diff
@@ -248,6 +252,35 @@ func (plugin *InterfaceConfigurator) ConfigureVPPInterface(iface *intf.Interface
 	return wasError
 }
 
+/**
+	Set rx-mode on specified VPP interface
```
VladoLavor: add to the comment info about what interface types can use what modes

plugins/defaultplugins/ifplugin/interface_config.go
```diff
@@ -248,6 +252,35 @@ func (plugin *InterfaceConfigurator) ConfigureVPPInterface(iface *intf.Interface
...
+		}
+	}
+	return nil
+}
+
+/**
+	Call concrete vpp API method for setting rx-mode
+ */
+func (plugin *InterfaceConfigurator) configRxMode(iface *intf.Interfaces_Interface, ifIdx uint32, rxModeSettings intf.Interfaces_Interface_RxModeSettings) error {
+	err := vppcalls.SetRxMode(ifIdx, rxModeSettings, plugin.Log, plugin.vppCh, nil)
```
VladoLavor: add stopwatch (instead of hardcoded nil)

measure.GetTimeLog(SwInterfaceSetRxMode{}, plugin.Stopwatch)

also in update/delete

plugins/defaultplugins/ifplugin/interface_config.go
jgloncak: done

plugins/defaultplugins/ifplugin/interface_config.go
jgloncak: done

plugins/defaultplugins/ifplugin/interface_config.go
jgloncak: is it also necessary to do delete of rx-mode?

PR #220, opened 2017-11-06
STN rules support

Batky:


plugins/defaultplugins/ifplugin/stn_config.go
```diff
@@ -0,0 +1,191 @@
...
+	return nil
+}
+
+// Check STN rule raw data
+func CheckStn(stnInput *model_stn.StnRule, index ifaceidx.SwIfIndex, log logging.Logger) (*vppcalls.StnRule, error) {
+	if stnInput == nil {
+		log.Infof("STN input is empty")
+		return nil, nil
+	}
+	if stnInput.Interface == "" {
```
VladoLavor: If the Interface and the IpAddress are required during both, Add and Delete, do not log it as an Info but return an error. The rule won't be configured in such a case which is an unwanted state.

plugins/defaultplugins/ifplugin/stn_config.go
```diff
@@ -0,0 +1,191 @@
...
+	plugin.Log.Infof("Creating new STN rule %v", rule)
+
+	// Check stn data
+	stnRule, err := CheckStn(rule, plugin.SwIfIndexes, plugin.Log)
+	if err != nil {
+		return err
+	}
+	plugin.Log.Debugf("adding STN rule: %+v", stnRule)
+	// Create and register new stn
+	errVppCall := vppcalls.AddStnRule(stnRule.IfaceIdx, &stnRule.IpAddress, plugin.Log, plugin.vppChan, measure.GetTimeLog(stn.StnAddDelRule{}, plugin.Stopwatch))
```
VladoLavor: stnRule.anything could throw nil pointer in current implementation if CheckStn returns nil, nil

plugins/defaultplugins/ifplugin/stn_config.go
```diff
@@ -0,0 +1,191 @@
...
+		&stn.StnAddDelRuleReply{},
+	}
+	err := plugin.vppChan.CheckMessageCompatibility(msgs...)
+	if err != nil {
+		plugin.Log.Error(err)
+	}
+	return err
+}
+
+func (plugin *StnConfigurator) Resync(stnRules []*model_stn.StnRule) error {
```
VladoLavor: move Resync func to ifplugin/data_resync

PR #237, opened 2017-11-10
Calling IP container VPP API

jgloncak:

### DIFF
--- a/plugins/defaultplugins/ifplugin/interface_config.go
+++ b/plugins/defaultplugins/ifplugin/interface_config.go
@@ -540,6 +540,8 @@ func (plugin *InterfaceConfigurator) modifyRxMode(ifIdx uint32, newIntf *intf.In
 		logging.Fields{"ifName": newIntf.Name, "rxMode old": oldRxMode, "rxMode new": newRxMode.RxMode}).
 		Debug("RX-mode modification for ", newIntf.Type, ".")
 	return err
+}
+
 func (plugin *InterfaceConfigurator) modifyContainerIPAddress(newConfig *intf.Interfaces_Interface, oldConfig *intf.Interfaces_Interface, ifIdx uint32) (int8, error) {
 	oldAddr, _, err := addrs.ParseIPWithPrefix(oldConfig.ContainerIpAddress)
 	if err != nil {
--- a/plugins/defaultplugins/ifplugin/interface_config.go
+++ b/plugins/defaultplugins/ifplugin/interface_config.go
@@ -50,7 +50,6 @@ import (
 	intf "github.com/ligato/vpp-agent/plugins/defaultplugins/ifplugin/model/interfaces"
 	"github.com/ligato/vpp-agent/plugins/defaultplugins/ifplugin/vppcalls"
 	"github.com/ligato/vpp-agent/plugins/govppmux"
-	"net"
 )
 
 const dummyMode = -1
@@ -219,15 +218,7 @@ func (plugin *InterfaceConfigurator) ConfigureVPPInterface(iface *intf.Interface
 	}
 
 	//configure container IP address
-	addr, isIpv6, err := addrs.ParseIPWithPrefix(iface.ContainerIpAddress)
-	if err != nil {
-		return err
-	}
-	err = vppcalls.AddContainerIP(ifIdx, addr, isIpv6, plugin.Log, plugin.vppCh,
-		measure.GetTimeLog(ip.IPContainerProxyAddDel{}, plugin.Stopwatch))
-	if nil != err {
-		wasError = err
-	}
+	plugin.addContainerIPAddress(iface, ifIdx)
 
 	// configure mtu
 	if iface.Type != intf.InterfaceType_VXLAN_TUNNEL {
@@ -265,21 +256,21 @@ func (plugin *InterfaceConfigurator) ConfigureVPPInterface(iface *intf.Interface
 }
 
 /**
-Set rx-mode on specified VPP interface
+	Set rx-mode on specified VPP interface
 
-Legend:
-P - polling
-I - interrupt
-A - adaptive
+	Legend:
+	P - polling
+	I - interrupt
+	A - adaptive
 
-Interfaces - supported modes:
-* tap interface - PIA
-* memory interface - PIA
-* vxlan tunnel - PIA
-* software loopback - PIA
-* ethernet csmad - P
-* af packet - PIA
-*/
+	Interfaces - supported modes:
+	* tap interface - PIA
+	* memory interface - PIA
+	* vxlan tunnel - PIA
+	* software loopback - PIA
+	* ethernet csmad - P
+	* af packet - PIA
+ */
 func (plugin *InterfaceConfigurator) configRxModeForInterface(iface *intf.Interfaces_Interface, ifIdx uint32) error {
 	rxModeSettings := iface.RxModeSettings
 	if rxModeSettings != nil {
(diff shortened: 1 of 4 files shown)


jgloncak:
In modification method of interface_config.go there **isn't called register and unregister Name method**.

It probably cause that if data are modified without delete/create attitude then in map there remain old value (e. g. ip container value is after modification still original which is possible to see in logs if we want to delete it)

plugins/defaultplugins/ifplugin/model/interfaces/interfaces.proto
```diff
@@ -20,6 +20,7 @@ message Interfaces {
         string phys_address = 5;
         uint32 mtu = 6;
         uint32 vrf = 7;
+        string container_ip_address = 8;
```
jgloncak: probably will be conflict once other pull request will be merged

plugins/defaultplugins/ifplugin/interface_config.go
```diff
@@ -215,6 +217,17 @@ func (plugin *InterfaceConfigurator) ConfigureVPPInterface(iface *intf.Interface
 		}
 	}
 
+	//configure container IP address
+	addr, isIpv6, err := addrs.ParseIPWithPrefix(iface.ContainerIpAddress)
```
VladoLavor: i think ParseIPWithPrefix returns an error if the input parameter is nil.  It makes container ip address a mandatory parameter and we do not want that. What I suggest:
- create some method called _SetContainerIP(containerIPAddr) error_ outside of ConfigureVPPInterface (the method is already long enough). Move all checks, parsing and vppcalls related to the container IP address to that method. 

- from CreateVPPInterface just call  SetContainerIP which sets the IPif exist, or does nothing. Do not return error if container IP is not set

plugins/defaultplugins/ifplugin/interface_config.go
```diff
@@ -522,6 +542,44 @@ func (plugin *InterfaceConfigurator) modifyRxMode(ifIdx uint32, newIntf *intf.In
...
+func (plugin *InterfaceConfigurator) modifyContainerIPAddress(newConfig *intf.Interfaces_Interface, oldConfig *intf.Interfaces_Interface, ifIdx uint32) (int8, error) {
+	oldAddr, _, err := addrs.ParseIPWithPrefix(oldConfig.ContainerIpAddress)
+	if err != nil {
+		return 1, err
+	}
+	newAddr, _, err := addrs.ParseIPWithPrefix(newConfig.ContainerIpAddress)
+	if err != nil {
+		return 1, err
+	}
+	toBeDeleted, toBeAdded := addrs.DiffAddr([]*net.IPNet{newAddr}, []*net.IPNet{oldAddr})
```
VladoLavor: I think this method is too complicated. Why you are calling DiffAddr on two slices where both contain just one item? You can just compare strings:

if oldConfig.ContainerIpAddress != newConfig.ContainerIpAddress {
       // convert to *net.IPNet and make the update
}

plugins/defaultplugins/ifplugin/interface_config.go
```diff
@@ -522,6 +542,44 @@ func (plugin *InterfaceConfigurator) modifyRxMode(ifIdx uint32, newIntf *intf.In
...
+
+	plugin.Log.Debug("Container IP address modification (delete): ", toBeDeleted)
+	plugin.Log.Debug("Container IP address modification (add): ", toBeAdded)
+
+	var wasError error
+	for i := range toBeDeleted {
```
VladoLavor: what happens if you call SetContainerIP if it already exists? I mean do we need to remove the old one?
And toBeDeleted will be always just one value (or none)

plugins/defaultplugins/ifplugin/interface_config.go
```diff
@@ -522,6 +542,44 @@ func (plugin *InterfaceConfigurator) modifyRxMode(ifIdx uint32, newIntf *intf.In
...
+		isIpv6, _ := addrs.IsIPv6(toBeDeleted[i].IP.String())
+		err := vppcalls.DelContainerIP(ifIdx, toBeDeleted[i], isIpv6, plugin.Log, plugin.vppCh,
+			measure.GetTimeLog(ip.IPContainerProxyAddDel{}, plugin.Stopwatch))
+		plugin.Log.Debug("Container IP address modification (delete) ", ifIdx, " ", toBeDeleted[i], " ", err)
+		if nil != err {
+			wasError = err
+		}
+	}
+
+	for i := range toBeAdded {
```
VladoLavor: again, toBeAdded is one item

plugins/defaultplugins/ifplugin/interface_config.go
```diff
@@ -571,6 +629,15 @@ func (plugin *InterfaceConfigurator) deleteVPPInterface(oldConfig *intf.Interfac
 	}
 
 	// let's try to do following even if previously error occurred
+	addr, isIpv6, err := addrs.ParseIPWithPrefix(oldConfig.ContainerIpAddress)
```
VladoLavor: do we need to do this if the interface will be removed?

plugins/defaultplugins/ifplugin/vppcalls/ip_container_vppcalls.go
```diff
@@ -0,0 +1,91 @@
...
+func AddContainerIP(ifIdx uint32, addr *net.IPNet, isIpv6 bool, log logging.Logger, vppChan *govppapi.Channel, timeLog measure.StopWatchEntry) error {
+	// IPContainerProxyAddDelReply time measurement
+	start := time.Now()
+	defer func() {
+		if timeLog != nil {
+			timeLog.LogTimeEntry(time.Since(start))
+		}
+	}()
+
+	req := prepareMessageForVpp(ifIdx, addr, isIpv6, 1)
```
VladoLavor: do not use magic numbers. Either define contants (add = 1, del = 0) or use boolean

plugins/defaultplugins/ifplugin/vppcalls/ip_container_vppcalls.go
```diff
@@ -0,0 +1,91 @@
...
+	} else {
+		req.IP = []byte(addr.IP.To16())
+		req.IsIP4 = 0
+	}
+	return req
+}
+
+func sendAndLogMessageForVpp(ifIdx uint32, req *ip.IPContainerProxyAddDel, logActionType string, log logging.Logger,  vppChan *govppapi.Channel) error {
+	log.WithFields(logging.Fields{"isIpv4": req.IsIP4, "prefix": req.Plen, "address": req.IP, "if_index": ifIdx}).
+		Debug("Container IP address ",logActionType,"ing...")
```
VladoLavor: interesting logging but it won't tell much, you should add interface index and IP address
the same on lines 85 & 87

plugins/defaultplugins/ifplugin/interface_config.go
jgloncak: as you wish

plugins/defaultplugins/ifplugin/interface_config.go
jgloncak: because I wasn't sure whether in future there won't be more container IP

plugins/defaultplugins/ifplugin/interface_config.go
jgloncak: you are right in both cases

plugins/defaultplugins/ifplugin/interface_config.go
jgloncak: right

plugins/defaultplugins/ifplugin/vppcalls/ip_container_vppcalls.go
jgloncak: ok

plugins/defaultplugins/ifplugin/vppcalls/ip_container_vppcalls.go
jgloncak: but it is part of logging - logging.Fields - address, if_index

plugins/defaultplugins/ifplugin/interface_config.go
jgloncak: I do not know where this setting (container IP address) is stored in VPP (whether it is part of interface configuration).
I would say that it isn't necessary to have it here, but I am not 100 % sure.

PR #277, opened 2017-11-21
Configuring routes waits for outgoing interface (VPP plugin)

jgloncak:
It is possible to specify set of routes for not existing
outgoing interface.
Once interface is created than this routes are automatically
configured on VPP.

In case of modification of cached route it is replaced with
new route (if there are same changes in priority or weight fields)

### DIFF
--- a/plugins/defaultplugins/l3plugin/route_config.go
+++ b/plugins/defaultplugins/l3plugin/route_config.go
@@ -53,7 +53,6 @@ type RouteConfigurator struct {
 
 // Init members (channels...) and start go routines.
 func (plugin *RouteConfigurator) Init() (err error) {
-	plugin.Log.SetLevel(logging.DebugLevel)
 	plugin.Log.Debug("Initializing L3 plugin")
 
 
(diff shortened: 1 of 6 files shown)


plugins/defaultplugins/l3plugin/route_config.go
```diff
@@ -44,14 +45,18 @@ type RouteConfigurator struct {
 	RouteIndexes  idxvpp.NameToIdxRW
 	RouteIndexSeq uint32
 	SwIfIndexes   ifaceidx.SwIfIndex
-	vppChan       *govppapi.Channel
-	Stopwatch     *measure.Stopwatch // timer used to measure and store time
+	RouteCachedIndex idxvpp.NameToIdxRW
```
VladoLavor: handle cache the route is deleted (remove the route from the cache if it is there)
also you should use custom type (for example RouteIndex instead of NameToIdxRW - see comments below)

plugins/defaultplugins/l3plugin/route_config.go
```diff
@@ -97,15 +115,50 @@ func (plugin *RouteConfigurator) ConfigureRoute(config *l3.StaticRoutes_Route, v
...
+	if outgoingIfName != "" {
+		_, _, existsNewOutgoing := plugin.SwIfIndexes.LookupIdx(outgoingIfName)
+		routeId := routeIdentifier(oldConfig.VrfId, oldConfig.DstIpAddr, oldConfig.NextHopAddr)
+		if existsNewOutgoing {
+			plugin.Log.Debugf("Route %s unregistered from cache.", routeId)
+			plugin.RouteCachedIndex.UnregisterName(routeId)
+		} else {
+			routeIdx,_, isRouteCached := plugin.RouteCachedIndex.LookupIdx(routeId)
+			if isRouteCached {
+				plugin.RouteCachedIndex.RegisterName(routeId, routeIdx, newConfig)
```
VladoLavor: missing plugin.RouteIndexSeq++

plugins/defaultplugins/l3plugin/route_config.go
```diff
@@ -211,3 +267,72 @@ func (plugin *RouteConfigurator) Close() error {
...
+	}
+}
+
+/**
+	recreateRoute calls delete and configure route.
+
+	This is type of workaround because when outgoing interface is deleted then it isn't possible to remove
+	associated routes. they stay in following state:
+	- oper-flags:drop
+	- routing section: unrosolved
```
VladoLavor: typo (unresolved)

jgloncak: thanks

plugins/defaultplugins/l3plugin/route_config.go
jgloncak: reason is that this is branch where route was found in cache. It means that it exists and that it already has assigned sequence number. It is in case where priority of weight (non key values) has been changed.

PR #327, opened 2017-12-07
Set ACL interfaces synchronously.

milanlenco:


plugins/defaultplugins/aclplugin/vppcalls/interfaces_vppcalls.go
```diff
@@ -14,144 +14,221 @@
...
+	interfaces []string
+	ingress    bool
+}
+
+// ACLInterfacesVppCalls aggregates vpp calls related to the IP ACL interfaces
+type ACLInterfacesVppCalls struct {
+	vppChan         *govppapi.Channel
+	swIfIndexes     ifaceidx.SwIfIndex
+	dumpIfaces      measure.StopWatchEntry
+	ifaceSetACLList measure.StopWatchEntry
```
ondrej-fabry: Variables `dumpIfaces` and `ifaceSetACLList` should include *stopwatch* in the name so it's clear when it's later used.

plugins/defaultplugins/aclplugin/vppcalls/interfaces_vppcalls.go
```diff
@@ -14,144 +14,221 @@
...
+		msg.SwIfIndex = index
+		msg.NInput = nInput
 
 		reply := &acl_api.ACLInterfaceSetACLListReply{}
-
-		err = vppChannel.SendRequest(req).ReceiveReply(reply)
+		err = acl.vppChan.SendRequest(msg).ReceiveReply(reply)
 		if err != nil {
-			return fmt.Errorf("failed to set interface %v to ACL %v", ingressInterface, aclIndex)
+			wasErr = err
```
ondrej-fabry: Should the loop continue even with error? Even if so, only the last error will be returned to the caller anyway. Also for large number of items that fail the time it will take to finish might be long.

plugins/defaultplugins/aclplugin/vppcalls/interfaces_vppcalls.go
```diff
@@ -14,144 +14,221 @@
...
-		log.Debugf("Interface %v set to ACL %v as ingress", ingressInterface, aclIndex)
 
-		// Log ACLInterfaceSetACLList time measurement results.
-		if timeLog != nil {
-			timeLog.LogTimeEntry(time.Since(start))
+		log.WithFields(logging.Fields{"SwIdx index": msg.SwIfIndex, "AclIdx": logicalReq.aclIndex}).Debug("Interface set to ACL")
+
+		// Log ACLInterfaceSetACLList time measurement results
+		if acl.ifaceSetACLList != nil {
+			acl.ifaceSetACLList.LogTimeEntry(time.Since(start))
```
ondrej-fabry: This will log time entries even if the call fails with error and give false overall measurements.

PR #443, opened 2018-02-07
Updating the path to Supervisord pid file

swatisehgal:
This issue was encountered when we use dev_vpp_agent image (which in my case I had build myself)


ondrej-fabry:
Closed due inactivity.

PR #460, opened 2018-02-21
docker/...: optimize & clean up Dockerfiles

unclejack:
This cleans up the Dockerfile for each of the two images.
It makes the following changes:
- installs most packages in the first phase
- removes the apt lists
- uses both g++ and gcc from the gcc7 ppa
- cleans up the formatting of the RUN commands

This reduced the size of the resulting image to 4.78 GB in my testing.

### DIFF
--- a/docker/prod_vpp_agent/Dockerfile
+++ b/docker/prod_vpp_agent/Dockerfile
@@ -2,12 +2,14 @@ FROM ubuntu:16.04
 
 MAINTAINER jozef.mak@pantheon.tech
 
-RUN apt-get update && apt-get install -y \
+RUN apt-get update \
+ && apt-get install -y \
         # general tools
         iproute2 iputils-ping inetutils-traceroute \
         # vpp requirements
         openssl python libapr1 libnuma1 \
-        supervisor
+        supervisor \
+ && rm -rf /var/lib/apt/lists/*
 
 # use /opt/vpp-agent/dev as the working directory
 RUN mkdir -p /opt/vpp-agent/dev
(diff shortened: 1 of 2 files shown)


docker/prod_vpp_agent/Dockerfile
```diff
@@ -2,12 +2,15 @@ FROM ubuntu:16.04
...
-RUN apt-get update && apt-get install -y \
+RUN apt-get update \
+ && apt-get install -y \
         # general tools
         iproute2 iputils-ping inetutils-traceroute \
         # vpp requirements
         openssl python libapr1 libnuma1 \
-        supervisor
+        supervisor \
+ && apt-get clean \
```
TrueFurby: Running `apt-get clean` is unnecessary, since the official Debian and Ubuntu images [automatically run `apt-get clean`](https://github.com/moby/moby/blob/03e2923e42446dbb830c654d0eec323a0b4ef02a/contrib/mkimage/debootstrap#L82-L105), so explicit invocation is not required.

ondrej-fabry: Running `apt-get clean` is unnecessary, since the official Debian and Ubuntu images [automatically run `apt-get clean`](https://github.com/moby/moby/blob/03e2923e42446dbb830c654d0eec323a0b4ef02a/contrib/mkimage/debootstrap#L82-L105), so explicit invocation is not required.

unclejack: @ondrej-fabry: I've removed that line from the Dockerfile.

PTAL

PR #620, opened 2018-05-18
Add unit tests for vpp/ifplugin/interface_state

deathbeam:
Signed-off-by: Tomas Slusny <tslusny@cisco.com>

### DIFF
--- a/plugins/defaultplugins/ifplugin/interface_state.go
+++ b/plugins/defaultplugins/ifplugin/interface_state.go
@@ -226,7 +226,7 @@ func (plugin *InterfaceStateUpdater) watchVPPNotifications(ctx context.Context) 
 
 		case swIdxDto := <-plugin.swIdxChan:
 			if swIdxDto.Del {
-				plugin.setIfStateDeleted(swIdxDto.Idx)
+				plugin.setIfStateDeleted(swIdxDto.Idx, swIdxDto.Name)
 			}
 			swIdxDto.Done()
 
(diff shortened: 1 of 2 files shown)


plugins/defaultplugins/ifplugin/interface_state.go
```diff
@@ -264,21 +264,26 @@ func (plugin *InterfaceStateUpdater) processIfStateNotification(notif *interface
 
 // getIfStateData returns interface state data structure for the specified interface index (creates it if it does not exist).
 // NOTE: plugin.ifStateData needs to be locked when calling this function!
-func (plugin *InterfaceStateUpdater) getIfStateData(swIfIndex uint32) (
+func (plugin *InterfaceStateUpdater) getIfStateData(swIfIndex uint32, isDel bool) (
```
VladoLavor: you could think about passing here ifName instead of some del flag. The problem here is that the ifName can't be Looked up (270) during delete, because the entry does not exist in the mapping. But the mapping is searched just for the name which can be passed directly to getIfStateData.

what I suggest:
1) remove isDel flag, add ifName string instead
2) remove LookupName, it is not needed anymore
3) all methods based on notifChan may obtain ifName using Lookup
4) method based on swIdxChan already contains ifName: swIdxDto.Name

please verify that everything works as intended, use vpp-agent-ctl -dump to see status data in the etcd

deathbeam: Updated

PR #643, opened 2018-05-28
Fix renamed configs for docker images

ondrej-fabry:


docker/dev/agent_start.py
```diff
@@ -13,7 +13,7 @@ def main():
         agent_start = os.environ['START_AGENT']
         if agent_start=='true' or agent_start == 'True':
             write_stdout('\nStarting vpp-agent...\n')
-            os.system("vpp-agent --etcdv3-config=/opt/vpp-agent/dev/etcd.conf --kafka-config=/opt/vpp-agent/dev/kafka.conf --default-plugins-config=/opt/vpp-agent/dev/vppplugin.conf --linuxplugin-config=/opt/vpp-agent/dev/linuxplugin.conf  --logs-config=/opt/vpp-agent/dev/logs.conf")
+            os.system("vpp-agent --config-dir=/opt/vpp-agent/dev")
```
VladoLavor: why were all the flags removed?

ondrej-fabry: So they can all load automatically from the config-dir, without having to set each one manually. This way any of them can be overridden with particular env variable, which previously cannot be done since generally flag overrides env vars.

PR #655, opened 2018-05-30
Added acl-plugin aclidx unit tests

pkotucek:


plugins/vpp/aclplugin/aclidx/aclidx_test.go
```diff
@@ -0,0 +1,89 @@
+package aclidx_test
```
VladoLavor: add licence header

PR #661, opened 2018-06-01
Added acl-plugin data resync unit tests

pkotucek:


plugins/vpp/aclplugin/vppdump/dump_vppcalls.go
```diff
@@ -66,6 +66,9 @@ func DumpIPACL(swIfIndices ifaceidx.SwIfIndex, log logging.Logger, vppChannel VP
 
 	// get all ACLs with IP ruleData
 	IPRuleACLs, _ := DumpIPAcls(log, vppChannel, stopwatch)
+	if len(IPRuleACLs) < 1 {
+		return ACLs, nil
```
ondrej-fabry: You could just simply return `nil` instead of `ACLs` here and then also move the `ACLs` var right before the last loop, since it's not used anywhere before it.

plugins/vpp/aclplugin/vppdump/dump_vppcalls.go
```diff
@@ -125,9 +128,9 @@ func DumpMACIPACL(swIfIndices ifaceidx.SwIfIndex, log logging.Logger, vppChannel
 	ruleMACIPData := make(map[ACLIdentifier][]*acl.AccessLists_Acl_Rule)
 
 	// get all ACLs with MACIP ruleData
-	MACIPRuleACLs, err := DumpMacIPAcls(log, vppChannel, stopwatch)
-	if err != nil {
-		return nil, err
+	MACIPRuleACLs, _ := DumpMacIPAcls(log, vppChannel, stopwatch)
```
ondrej-fabry: Why ignore error here?

plugins/vpp/aclplugin/vppdump/dump_vppcalls.go
```diff
@@ -125,9 +128,9 @@ func DumpMACIPACL(swIfIndices ifaceidx.SwIfIndex, log logging.Logger, vppChannel
...
+	if len(MACIPRuleACLs) < 1 {
+		return ACLs, nil
```
ondrej-fabry: Same thing here, just do `return nil, nil`. This makes it clear instantly when looking at this line that slice is empty without going back to check if there is any append before this.

plugins/vpp/aclplugin/acl_config_test.go
```diff
@@ -421,29 +421,33 @@ func TestResolveDeletedInterface(t *testing.T) {
...
-	ctx.MockVpp.MockReply(&acl_api.ACLPluginGetVersionReply{})
-	err = plugin.Init(logging.ForPlugin("test-log", logrus.NewLogRegistry()), connection, ifIndexes, false)
+	// Configurator
+	ctx.MockVpp.MockReply(&acl_api.ACLPluginGetVersionReply{1, 0})
+	plugin := &aclplugin.ACLConfigurator{}
+	err = plugin.Init(log, connection, ifIndexes, false)
 	Expect(err).To(BeNil())
 
-	return ctx, connection, plugin
+	return ctx, connection, plugin, log
```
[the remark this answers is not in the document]
pkotucek: it is used Resync tests.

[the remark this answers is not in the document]
ondrej-fabry: Oh, I see. However that is not the correct. None of the other resyncs take logger and it's generally not very idiomatic to pass around loggers via parameters. 

The `Resync` for ACLs is a method for ACL configurator which has it's own logger. Please get rid of this log parameter in `Resync` and replace all its uses with `plugin.log.X`.

plugins/vpp/aclplugin/data_resync_test.go
```diff
@@ -0,0 +1,419 @@
...
+		},
+		{
+			Id:   1005,
+			Ping: false,
+			Message: &acl_api.MacipACLAddReply{},
+		},
+	}))
+
+	err := plugin.Resync(acls)
+	Expect(err).To(BeNil())
```
VladoLavor: please create two new methods - for l2AclIndexes and l3l4AclIndexes - in acl configurator which allow to lookup in-memory mapping  ([example](https://github.com/ligato/vpp-agent/blob/master/plugins/defaultplugins/ifplugin/interface_config.go#L144-L147)) and check also mapping state in every test here (same as [here](https://github.com/ligato/vpp-agent/blob/master/plugins/defaultplugins/ifplugin/interface_config_test.go#L220-L222)). Plugin internal state is essential and it would be good if it could be tested also in acl_configurator_test

plugins/vpp/aclplugin/data_resync_test.go
```diff
@@ -0,0 +1,489 @@
...
+				NInput:    1,
+				Acls:      []uint32{0, 2},
+			},
+		},
+		{
+			Id:   1013,
+			Ping: true,
+			Message: &acl_api.MacipACLDetails{
+				ACLIndex: 0,
+				Tag:      []byte{'a', 'c', 'l', '4'},
```
ondrej-fabry: You can convert from `string` to `[]byte` using normal brackets: `[]byte("acl4")`.

plugins/vpp/aclplugin/data_resync_test.go
```diff
@@ -0,0 +1,489 @@
...
+	defer aclTestTeardown(connection, plugin)
+
+	ctx.MockVpp.RegisterBinAPITypes(acl_api.Types)
+	ctx.MockVpp.MockReplyHandler(vppMockHandler(ctx.MockVpp, []*vppReplyMock{
+		{
+			Id:   1011,
+			Ping: true,
+			Message: &acl_api.ACLDetails{
+				ACLIndex: 0,
+				Tag:      []byte{'a', 'c', 'l', '3'},
```
ondrej-fabry: Same here. And few other places to. Use Ctrl+Shift+F and search for `[]byte{`.

PR #691, opened 2018-06-15
Tolerate removal of AUTO-TAP from outside of the agent

pkotucek:
Changed the way how linux plugin reacts on interface changes in VPP.
Currently it is done the same way as VPP reacts on linux interface
changes - there is a new channel vppIfIndexesWatchChan which is used to
notify linux plugin about interface changes in VPP.
Side effect is removal of unused LinuxInterfaceStateNotification and
LinuxInterfaceStateUpdater.


pkotucek:
It has been changed to standard way as for other interfaces, i.e. in linux intf there are Create/Delete resolve methods to resolve interface changes in vpp. Reverse direction (vpp resolves changes in linux intf) has been done already.

plugins/linux/ifplugin/interface_state.go
```diff
@@ -1,134 +0,0 @@
-// Copyright (c) 2018 Cisco and/or its affiliates.
```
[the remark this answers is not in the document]
ondrej-fabry: Agreed.

plugins/linux/ifplugin/interface_config.go
```diff
@@ -992,3 +969,60 @@ func addressExists(configured []netlink.Addr, provided *net.IPNet) bool {
...
+	plugin.log.Infof("Linux IF configurator: resolve deleted vpp interface %v", ifConfigMetaData)
+
+	if ifConfigMetaData.Type != vppIf.InterfaceType_TAP_INTERFACE {
+		return nil
+	}
+
+	var hostIfName string
+	if ifConfigMetaData.Tap != nil && ifConfigMetaData.Tap.HostIfName != "" {
+		hostIfName = ifConfigMetaData.Tap.HostIfName
+	} else if ifConfigMetaData != nil && ifConfigMetaData.Name != "" {
```
ondrej-fabry: This check is pointless because if the `ifConfigMetaData` will be `nil`, the program will crash on line `996` due accessing its fields.

plugins/linux/ifplugin/interface_config.go
```diff
@@ -992,3 +969,60 @@ func addressExists(configured []netlink.Addr, provided *net.IPNet) bool {
...
+// ResolveCreatedInterface resolves a new vpp interfaces
+func (plugin *LinuxInterfaceConfigurator) ResolveCreatedVPPInterface(ifConfigMetaData *vppIf.Interfaces_Interface) error {
+	plugin.log.Infof("Linux IF configurator: resolve created vpp interface %v", ifConfigMetaData)
+
+	if ifConfigMetaData.Type != vppIf.InterfaceType_TAP_INTERFACE {
+		return nil
+	}
+
+	var hostIfName string
+	if ifConfigMetaData.Tap != nil && ifConfigMetaData.Tap.HostIfName != "" {
```
ondrej-fabry: You can use `ifConfigMetaData.GetTap().GetHostIfName()` to avoid `nil` check.

plugins/linux/ifplugin/interface_config.go
```diff
@@ -320,13 +331,36 @@ func (plugin *LinuxInterfaceConfigurator) configureVethInterface(ifConfig, peerC
 
 // Validate and apply linux TAP configuration to the interface. The interface is not created here, it is added
 // to the default namespace when it's VPP end is configured
-func (plugin *LinuxInterfaceConfigurator) configureTapInterface(ifConfig *LinuxInterfaceConfig) error {
+//func (plugin *LinuxInterfaceConfigurator) configureTapInterface(ifConfig *LinuxInterfaceConfig) error {
```
ondrej-fabry: Forgotten line?

plugins/linux/ifplugin/interface_config.go
```diff
@@ -260,6 +264,13 @@ func (plugin *LinuxInterfaceConfigurator) DeleteLinuxInterface(linuxIf *interfac
 		peerConfig = oldConfig.peer
 	}
 
+	if _, _, exists := plugin.ifCachedConfigs.LookupIdx(linuxIf.HostIfName); exists {
+		// Unregister TAP from the in-memory map
+		plugin.log.Infof("Removing Linux Tap interface %v from cache", linuxIf.HostIfName)
```
ondrej-fabry: It is not certain here that this interface is TAP, is it? If the `ifCachedConfigs` is a cache for TAP interfaces only and those are registered in `configureTapInterface` function, then they should be unregistered inside `deleteTapInterface` function in which we are sure we work with TAP interface. Otherwise situation might arise where we delete cached TAP interface if they for some reason have same `HostIfName`.

plugins/linux/ifplugin/interface_config.go
pkotucek: :-) you are right. I will change it

plugins/linux/linuxplugin_init.go
```diff
@@ -45,23 +47,21 @@ type Plugin struct {
 	arpConfigurator   *l3plugin.LinuxArpConfigurator
 	routeConfigurator *l3plugin.LinuxRouteConfigurator
 
-	// State updaters
-	ifLinuxStateUpdater *ifplugin.LinuxInterfaceStateUpdater
-
 	// Shared indexes
-	ifIndexes ifaceidx.LinuxIfIndexRW
+	ifIndexes   ifaceidx.LinuxIfIndexRW
+	swIfIndexes ifaceVPP.SwIfIndex
```
ondrej-fabry: It might be more clear to reader if this was named `vppIfIndexes`.

cmd/vpp-agent-ctl/data_cmd.go
```diff
@@ -314,7 +314,7 @@ func (ctl *VppAgentCtl) createTap() {
 				//	InterfaceWithIP: "memif1",
 				//},
 				Tap: &interfaces.Interfaces_Interface_Tap{
-					HostIfName: "tap1",
+					HostIfName: "tap-host",
```
ondrej-fabry: This change doesn't affect any other commands, like Vlado mentioned earlier?

pkotucek: Without this change it doesn't work, as hostname for tap and ltap are not the same and so tap is not configured correctly. How can I verify whether it affect other commands?

ondrej-fabry: Actually, the name of the interface `tap1` stayed the same, so hopefully nothing else will be affected since most other referings are through the `Name` and not `HostIfName`.

PR #697, opened 2018-06-19
Correctly set afpacket mac

deathbeam:
Instead of using VPP call to set afpacket hardware address after
creation, do it when creating afpacket interface using afpacket-specific
calls. This prevents afpacket setting its own MAC address to bound veth.

Signed-off-by: Tomas Slusny <tslusny@cisco.com>


deathbeam:
Added test case

PR #719, opened 2018-07-02
Some tests for TAP interfaces started failing

pkotucek:
Some tests for TAP interfaces started failing recently. Probably after
this change: https://github.com/ligato/vpp-agent/pull/691

### DIFF
--- a/plugins/linux/ifplugin/interface_config.go
+++ b/plugins/linux/ifplugin/interface_config.go
@@ -156,7 +156,14 @@ func (plugin *LinuxInterfaceConfigurator) ConfigureLinuxInterface(linuxIf *inter
 
 		return plugin.configureVethInterface(ifConfig, peerConfig)
 	case interfaces.LinuxInterfaces_AUTO_TAP:
-		return plugin.configureTapInterface(linuxIf.HostIfName, linuxIf)
+		var hostIfName string = ""
+		if linuxIf.Tap != nil  {
+			hostIfName = linuxIf.GetTap().GetTempIfName()
+		}
+		if hostIfName == "" {
+			hostIfName = linuxIf.GetHostIfName()
+		}
+		return plugin.configureTapInterface(hostIfName, linuxIf)
 	default:
 		return fmt.Errorf("unknown linux interface type: %v", linuxIf.Type)
 	}
@@ -347,7 +354,7 @@ func (plugin *LinuxInterfaceConfigurator) configureTapInterface(hostIfName strin
 				Data:  ifConfig.config,
 			})
 			plugin.pIfCachedConfigSeq++
-			plugin.log.Infof("Creating new Linux Tap interface  %v configuration entry %v", hostIfName, ifConfig.config)
+			plugin.log.Infof("Creating new Linux Tap interface %v configuration entry %v", hostIfName, ifConfig.config)
 		} else {
 			plugin.log.Infof("There is no Linux Tap configuration entry for interface %v", hostIfName)
 			return nil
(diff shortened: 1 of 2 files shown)


plugins/linux/ifplugin/interface_config.go
```diff
@@ -993,7 +1000,14 @@ func (plugin *LinuxInterfaceConfigurator) ResolveCreatedVPPInterface(ifConfigMet
 		plugin.log.Warn("Unable to resolve created VPP interfaces, incomplete Linux TAP configuration data")
 		return nil
 	}
-	return plugin.configureTapInterface(hostIfName, nil)
+
+	var linuxIf *interfaces.LinuxInterfaces_Interface = nil
```
ondrej-fabry: No need to initialize variables, they always start with zero value.

plugins/linux/ifplugin/interface_config.go
```diff
@@ -156,7 +156,14 @@ func (plugin *LinuxInterfaceConfigurator) ConfigureLinuxInterface(linuxIf *inter
 
 		return plugin.configureVethInterface(ifConfig, peerConfig)
 	case interfaces.LinuxInterfaces_AUTO_TAP:
-		return plugin.configureTapInterface(linuxIf.HostIfName, linuxIf)
+		var hostIfName string = ""
```
ondrej-fabry: No need to initialize.

plugins/linux/ifplugin/interface_config.go
```diff
@@ -156,7 +156,14 @@ func (plugin *LinuxInterfaceConfigurator) ConfigureLinuxInterface(linuxIf *inter
...
+		if linuxIf.Tap != nil  {
+			hostIfName = linuxIf.GetTap().GetTempIfName()
+		}
+		if hostIfName == "" {
```
ondrej-fabry: You can use `else` case instead of checking if the `hostIfName` is empty.

pkotucek: But if tempIfName is empty the hostIfName will be empty even tap's hostIfName is defined.

VladoLavor: if linuxIf.Tap != nil && linuxIf.Tap.TempIfName != "" {
    return plugin.configureTapInterface(linuxIf.Tap.TempIfName, linuxIf)
} else {
    return plugin.configureTapInterface(linuxIf.HostIfName, linuxIf)
}

HostIfName is always set. In this case you can alse remove the first parameter (temp/host name can be read from linuxIf)

plugins/linux/ifplugin/interface_config_test.go
```diff
@@ -461,6 +461,43 @@ func TestLinuxConfiguratorAddVethPairError(t *testing.T) {
...
+	// Linux/namespace calls
+	nsMock.When("IsNamespaceAvailable").ThenReturn(true)
+
+	data := getTapInterface("tap1", "", "TempIfName")
+	err := plugin.ConfigureLinuxInterface(data)
+	Expect(err).ShouldNot(HaveOccurred())
+	Expect(plugin.GetLinuxInterfaceIndexes().GetMapping().ListNames()).To(HaveLen(0))
+	Expect(plugin.GetCachedLinuxIfIndexes().GetMapping().ListNames()).To(HaveLen(1))
+	// Verify registration
+	_, _, found := plugin.GetCachedLinuxIfIndexes().LookupIdx("TempIfName")
```
VladoLavor: please add metadata nil check

PR #722, opened 2018-07-03
Fix tests with MockHandler

ondrej-fabry:


plugins/linux/nsplugin/microservices.go
```diff
@@ -228,7 +228,10 @@ func (plugin *NsHandler) processTerminatedMicroservice(nsMgmtCtx *NamespaceMgmtC
 // trackMicroservices is running in the background and maintains a map of microservice labels to container info.
 func (plugin *NsHandler) trackMicroservices(ctx context.Context) {
 	plugin.wg.Add(1)
-	defer plugin.wg.Done()
+	defer func() {
+		plugin.wg.Done()
+		plugin.log.Debugf("trackMicroservices done")
```
VladoLavor: I would say something like 'Microservice tracking ended'

PR #738, opened 2018-07-12
Fix for failing SR robot tests

fgschwan:


fgschwan:
@ondrej-fabry @VladoLavor ,
Hi, this is small fix for big SR problems. It would be great if it can make it to 1.5 release. Thx.

PR #739, opened 2018-07-13
Fix for upcomming big SR API change in VPP

fgschwan:
Change in VPP is only 4 commits ahead of currently used VPP.  Therefore SR functionality would probably break when VPP will be updated in vpp agent.


fgschwan:
@ondrej-fabry 
@VladoLavor 
This is for future updates of reference to used VPP, so it don't have to go inside 1.5 release. Current dev branch (that will be 1.5 release) has working SR functionality.

plugins/vpp/srplugin/vppcalls/srv6.go
```diff
@@ -380,23 +387,31 @@ func (calls *srv6Calls) modPolicy(operation uint8, bindingSid net.IP, policy *sr
 	return nil
 }
 
-func (calls *srv6Calls) convertNextSidList(nextSidList []string) (uint8, []sr.SrIP6Address, error) {
-	segments := make([]sr.SrIP6Address, 0)
-	for _, sid := range nextSidList {
+func (calls *srv6Calls) convertPolicySegment(policySegment *srv6.PolicySegment) (*sr.Srv6SidList, error) {
+	segments := make([]sr.Srv6Sid, 0)
```
ondrej-fabry: This doesn't need to be so verbose if you set length to zero, `var segments []sr.Srv6Sid` is enough.

Makefile
```diff
@@ -136,6 +136,7 @@ generate-binapi: get-binapi-generators
 	cd plugins/vpp/binapi/vxlan && pkgreflect
 	@echo "=> applying fix patch"
 	patch -p1 -i plugins/vpp/binapi/fixapi.patch
+	patch -p1 -i plugins/vpp/binapi/fixapi-sr.patch
```
ondrej-fabry: We should probably just use `*.patch` here to avoid future additions and support splitting patches into files.

plugins/vpp/srplugin/vppcalls/srv6.go
```diff
@@ -380,23 +387,31 @@ func (calls *srv6Calls) modPolicy(operation uint8, bindingSid net.IP, policy *sr
...
+	for _, sid := range policySegment.Segments {
 		// parse to IPv6 address
 		parserSid, err := parseIPv6(sid)
 		if err != nil {
-			return 0, []sr.SrIP6Address{}, err
+			return &sr.Srv6SidList{
```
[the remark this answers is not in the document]
fgschwan: fixed in 1f9e360

plugins/vpp/srplugin/vppcalls/srv6.go
fgschwan: fixed in 1f9e360

Makefile
fgschwan: I agree, but that code that you commited didn't do the trick in this case. I changed it to "find -exec" version, but with the same goal. See the fix in 3543e48.

PR #747, opened 2018-07-18
Support for inter-VRF routes

rastislavszabo:
This MR introduces inter-VRF routing (routes that refer to a different VRF).


plugins/vpp/l3plugin/route_utils.go
```diff
@@ -89,8 +97,11 @@ func TransformRoute(routeInput *l3.StaticRoutes_Route, swIndex uint32, log loggi
 		return nil, nil
 	}
 	if routeInput.DstIpAddr == "" {
-		log.Infof("Route does not contain destination address")
-		return nil, nil
+		if routeInput.Type != l3.StaticRoutes_Route_INTER_VRF {
```
VladoLavor: destination address should be set for every route type

plugins/vpp/model/l3/l3.proto
```diff
@@ -5,15 +5,26 @@ package l3;
...
+        enum RouteType {
+            INTRA_VRF = 0;                   /* Forwarding is being done in the specified vrf_id only, or according to
+                                                the specified outgoing interface. */
+            INTER_VRF = 1;                   /* Forwarding is being done by lookup into a different VRF,
+                                                specified as via_vrf_id field. In case of these routes, the outgoing
+                                                interface should not be specified. The next hop IP address
+                                                does not have to be specified either, in that case VPP does full
+                                                recursive lookup in the via_vrf_id VRF. */
+        }
+        RouteType type = 1;
```
ondrej-fabry: I don't think we should change the identification for fields. That's how some implementations for proto identify and decode the data.

rastislavszabo: Makes sense if we only want to do backward-compatible changes in the proto files. This change makes it non-backward compatible. I can make it the member number 10 if you prefer that.

Do we have any guidelines on when we do only backward-compatible changes and when we are allowed to do non-backward compatible changes?

ondrej-fabry: Nothing written in repo AFAIK.. just what we discussed verbally maybe with other teams.

PR #756, opened 2018-07-20
l3plugin static route dump should return all paths

pkotucek:
The l3plugin static route dump currently returns only the first hop in given route.
Fix returns all hops in route

### DIFF
--- a/plugins/vpp/l3plugin/vppdump/dump_vppcalls.go
+++ b/plugins/vpp/l3plugin/vppdump/dump_vppcalls.go
@@ -30,7 +30,7 @@ import (
 )
 
 // DumpStaticRoutes dumps l3 routes from VPP and fills them into the provided static route map.
-func DumpStaticRoutes(log logging.Logger, vppChan govppapi.Channel, timeLog measure.StopWatchEntry) ([]*vppcalls.Route, error) {
+func DumpStaticRoutes(log logging.Logger, vppChan govppapi.Channel, timeLog measure.StopWatchEntry) ([][]*vppcalls.Route, error) {
 	// IPFibDump time measurement
 	start := time.Now()
 	defer func() {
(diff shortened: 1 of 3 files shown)


plugins/vpp/l3plugin/vppdump/dump_vppcalls.go
```diff
@@ -101,33 +101,35 @@ func dumpStaticRouteIPDetails(tableID uint32, tableName []byte, address []byte,
 		ipAddr = fmt.Sprintf("%s/%d", net.IP(address[:4]).To4().String(), uint32(prefixLen))
 	}
 
-	rt := &vppcalls.Route{}
+	rts := []*vppcalls.Route{}
```
ondrej-fabry: No need to initialize the slice. You can just keep it nil: `var rts []*vppcalls.Router`.

The `append` function will take care of initializing it.

plugins/vpp/l3plugin/vppdump/dump_vppcalls.go
```diff
@@ -39,7 +39,7 @@ func DumpStaticRoutes(log logging.Logger, vppChan govppapi.Channel, timeLog meas
 		}
 	}()
 
-	var routes []*vppcalls.Route
+	var routes [][]*vppcalls.Route
```
VladoLavor: this solution seems not very readable for me, also you will have a lot of duplicate fields in the resulting list. 
The destination address, table ID (Vrf) and table name are the same for every path. Path fields are next-hop, interface, weight and preference. I'd prefer to create a new struct Path (for example) with all these fields, and add a list of Path objects to Route.

ondrej-fabry: I agree. Having simple struct even with single slice field will be much more readable because it can be named and documented more easily.

VladoLavor:
fixed here: https://github.com/ligato/vpp-agent/pull/766

PR #777, opened 2018-08-06
[contiv-1807] VPP ver. change to contiv-1807-stable with bin API change in NAT

rastislavszabo:


docker/dev/Dockerfile
```diff
@@ -26,7 +26,7 @@ COPY Makefile vpp.env ./
 
 RUN cp -r plugins/vpp/binapi /tmp/orig_binapi \
  && make generate-binapi \
- && diff -r plugins/vpp/binapi /tmp/orig_binapi
+ && diff -r -N -q plugins/vpp/binapi /tmp/orig_binapi
```
ondrej-fabry: This does not help with what you probably intended. The `patch` command used in Makefile target `verify-binapi` needs `--no-backup-if-mismatch` flag to not produce `*.orig` files on mismatch. Then the verification should succeed.

vendor/git.fd.io/govpp.git/cmd/binapi-generator/generator.go
```diff
@@ -540,7 +540,8 @@ func translateVppType(ctx *context, vppType string, isArray bool) string {
 		return typ
 	}
 
-	panic(fmt.Sprintf("Unknown VPP type %s", vppType))
+	//panic(fmt.Sprintf("Unknown VPP type %s", vppType))
+	return "uint8"
```
ondrej-fabry: It seems like `enums` were finally introduced in VPP's API. This will need to be added to GoVPP as well.

plugins/vpp/ifplugin/data_resync.go
```diff
@@ -600,8 +600,8 @@ func (plugin *NatConfigurator) resolveMappings(nbDNatConfig *nat.Nat44DNat_DNatC
 				for _, nbLocal := range nbMapping.LocalIps {
 					var found bool
 					for _, vppLocal := range vppLbMapping.LocalIps {
-						if nbLocal.LocalIp == vppLocal.LocalIp || nbLocal.LocalPort == vppLocal.LocalPort ||
```
ondrej-fabry: This seems like bug.. right?

Makefile
```diff
@@ -144,7 +144,7 @@ generate-binapi: get-binapi-generators
 	cd plugins/vpp/binapi/vpe && pkgreflect
 	cd plugins/vpp/binapi/vxlan && pkgreflect
 	@echo "=> applying fix patches"
-	find plugins/vpp/binapi -maxdepth 1 -type f -name '*.patch' -exec patch -p1 -i {} \;
+	find plugins/vpp/binapi -maxdepth 1 -type f -name '*.patch' -exec patch -p1 -i --no-backup-if-mismatch {} \;
```
ondrej-fabry: `patch --no-backup-if-mismatch -p1 -i ...`

plugins/vpp/ifplugin/data_resync.go
rastislavszabo: yep, bug. Please fix in pantheon-dev too

PR #784, opened 2018-08-09
Do not hide error in parsing telemetry config

lukasmacko:


ondrej-fabry:
Seems like linux is using same invalid logic for checking config.

lukasmacko:
yes, the linux plugin is having the same issue.

PR #816, opened 2018-08-22
Update dependencies

mestery:
This updates all the dependencies to the very latest, and is part of
the work to add CRD support to contiv-vpp.

Note that this is dependent on an unmerged patch for cn-infra, which
can be found here [1].

[1] https://github.com/ligato/cn-infra/pull/328

Signed-off-by: Kyle Mestery <mestery@mestery.com>


mestery:
@ondrej-fabry this one is now against the pantheon-dev branch of vpp-agent.

ondrej-fabry:
You build fails on Travis, because you updated govpp as well, which has changed some API. Maybe keep this on the previous version so you don't have to deal with that. We periodically update cn-infra & govpp so that is not so needed.

mestery:
I suggest we merge a few of the Dependabot items which should pass CI before we merge this one in. I'll rebase this one once those merge.

ondrej-fabry:
@mestery ready for rebase! 👍

Gopkg.toml
```diff
@@ -6,12 +6,18 @@
 required = [
     "git.fd.io/govpp.git/cmd/binapi-generator",
     "github.com/ungerik/pkgreflect",
-    "github.com/gogo/protobuf/protoc-gen-gogo"
+    "github.com/gogo/protobuf/protoc-gen-gogo",
+    "github.com/Shopify/sarama",
+    "github.com/bsm/sarama-cluster",
+    "github.com/coreos/etcd",
```
ondrej-fabry: Many of these don't need to be listed in `required`, since they are imported as direct or transitive dependency. Mainly only some programs which cannot be imported, but are used for code generation etc. should be here.

Gopkg.toml
```diff
@@ -30,10 +31,6 @@ required = [
   name = "github.com/ghodss/yaml"
   version = "1.0.0"
 
-[[constraint]]
```
ondrej-fabry: Why was this removed from here? It's direct dependency we use in vpp-agent.

Gopkg.toml
```diff
@@ -101,7 +106,3 @@ required = [
   [[prune.project]]
     name = "github.com/ligato/cn-infra"
     unused-packages = false
-
-  [[prune.project]]
```
ondrej-fabry: This served as a workaround for `dep check` failure, which actually occurred in the last build on this PR:

```sh
dep check
# vendor is out of sync:
github.com/go-errors/errors: hash of vendored tree not equal to digest in Gopkg.lock
```

PR #867, opened 2018-09-07
Fix protobuf imports

ondrej-fabry:


Makefile
```diff
@@ -105,16 +105,8 @@ get-proto-generators:
...
-	cd plugins/vpp/aclplugin && go generate
-	cd plugins/vpp/ifplugin && go generate
-	cd plugins/vpp/ipsecplugin && go generate
-	cd plugins/vpp/l2plugin && go generate
-	cd plugins/vpp/l3plugin && go generate
-	cd plugins/vpp/l4plugin && go generate
-	cd plugins/vpp/rpc && go generate
-	cd plugins/vpp/srplugin && go generate
+	cd plugins/linux/model && go generate
+	cd plugins/vpp/model && go generate
```
milanlenco: With the refactor we will be generating code inside plugins anyway (descriptor boiler-plate) - so this make target will grow back again.

ondrej-fabry: That will most likely be separated into new make target, something like `generate-descriptors`. We have one for proto models (this one) and for binapi as separate already.

PR #892, opened 2018-09-20
Linux Plugin v2 (with KVScheduler)

milanlenco:


plugins/linuxv2/model/l3/keys.go
```diff
@@ -0,0 +1,144 @@
...
+	staticLinkLocalRouteKeyPrefix = "linux/link-local-route/"
+
+	// staticLinkLocalRouteKeyTemplate is a template for key derived from link-local route.
+	staticLinkLocalRouteKeyTemplate = staticLinkLocalRouteKeyPrefix + staticRouteKeySuffix
+)
+
+/* ARP */
+
+// StaticArpKeyPrefix returns the prefix used in ETCD to store config for Linux static ARPs
+func StaticArpKeyPrefix() string {
```
ondrej-fabry: In the VPP plugin we have already got rid of these *helper functions* which return constatns. All prefixes should be exported so they can be used directly and leave only functions which concatenate keys. 

Aim for less code ☀️

plugins/linuxv2/l3plugin/l3plugin.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+// Config holds the nsplugin configuration.
+type Config struct {
+	Stopwatch bool `json:"stopwatch"`
+	Disabled  bool `json:"disabled"`
+}
+
+// Init initializes and registers descriptors for Linux ARPs and Routes.
+func (plugin *L3Plugin) Init() error {
```
ondrej-fabry: `plugin` -> `p`

Let's keep all method receivers short, ideally just a single character. It's generally the most repeated variable for methods and such variables should not be so verbose.

plugins/linuxv2/model/interfaces/interfaces.proto
```diff
@@ -0,0 +1,33 @@
+syntax = "proto3";
+
+package interfaces;
+
+import "github.com/ligato/vpp-agent/plugins/linuxv2/model/namespace/namespace.proto";
+
+enum LinuxInterfaceType {               /* Available interface types */
```
ondrej-fabry: It would be much cleaner to have this defined inside `LinuxInterface` message. The generated type will also be easier to read. `interfaces.LinuxInterface_Type` vs. `interfaces.LinuxInterfaceType`.

plugins/linuxv2/model/namespace/namespace.proto
```diff
@@ -0,0 +1,19 @@
+syntax = "proto3";
+
+package namespace;
+
+message Namespace {              /* Linux network namespace */
+    enum NamespaceType {
+        PID_REF_NS = 0;          /* namespace of a given process */
```
ondrej-fabry: Since we are doing rewrite of the API.. we should consider defining these network namespace types same way as they are [defined in the linux kernel](https://github.com/torvalds/linux/blob/6f0d349d922ba44e4348a17a78ea51b7135965b1/include/uapi/linux/net_namespace.h#L13-L19).

```c
enum {
	NETNSA_NONE,
	NETNSA_NSID,
	NETNSA_PID,
	NETNSA_FD,
    //...
};
```

plugins/linuxv2/model/namespace/namespace.proto
```diff
@@ -0,0 +1,19 @@
+syntax = "proto3";
+
+package namespace;
+
+message Namespace {              /* Linux network namespace */
+    enum NamespaceType {
```
ondrej-fabry: `namespace.Namespace_NamespaceType` does not look very clean.

I would suggest to keep just `Type` for the enum.

plugins/linuxv2/model/namespace/namespace.proto
```diff
@@ -0,0 +1,19 @@
+syntax = "proto3";
+
+package namespace;
+
+message Namespace {              /* Linux network namespace */
```
ondrej-fabry: This is debatable.. All other linux models now have `Linux` prefix and this one does not, also it should probably be more clear that it's network namespace since there are other types of namespaces in linux. Going with `LinuxNetworkNamespace` might be too long, but we could shorten it to: `LinuxNetNamespace`, this will also remove the stutter: `namespace.Namespace..`.

plugins/linuxv2/model/namespace/namespace.proto
```diff
@@ -0,0 +1,19 @@
...
+    enum NamespaceType {
+        PID_REF_NS = 0;          /* namespace of a given process */
+        MICROSERVICE_REF_NS = 1; /* namespace of a docker container running given microservice */
+        NAMED_NS = 2;            /* named namespace */
+        FILE_REF_NS = 3;         /* namespace referenced by a file handle*/
+    }
```
ondrej-fabry: Enum definitions which are defined inside message should be defined just before first use as field in that message. So here, it should be just before the `type`.

plugins/linuxv2/model/namespace/namespace.proto
```diff
@@ -0,0 +1,19 @@
...
+    string name = 1;             /* logical name (mandatory) - also used as the physical name if type==NAMED_NS */
+    NamespaceType type = 2;      /* type of the namespace reference (mandatory) */
+
+    /* based on the type, one of these must be defined: */
```
ondrej-fabry: We should consider using `oneof` field for these.

plugins/linuxv2/model/interfaces/interfaces.proto
```diff
@@ -0,0 +1,33 @@
...
+};
+
+message LinuxInterface {
+    string name = 1;                    /* Logical interface name unique across all configured interfaces (mandatory) */
+    LinuxInterfaceType type = 2;        /* Interface type (mandatory) */
+    bool enabled = 3;
+    repeated string ip_addresses = 4;   /* IP address in format <ipAddress>/<ipPrefix> */
+    string phys_address = 5;            /* MAC address */
+    uint32 mtu = 6;                     /* Maximum transmission unit value */
+    string host_if_name = 7;            /* Name of the interface in the host OS. If not set, the host name
```
ondrej-fabry: Couldn't we get rid of this and use just `name`?

plugins/linuxv2/model/interfaces/interfaces.proto
```diff
@@ -0,0 +1,33 @@
...
+                                           is the same as interface name */
+
+    namespace.Namespace namespace = 8;
+
+    message Veth {
+        string peer_if_name = 1;         /* Name of the peer, i.e. other end of the linux veth (mandatory for VETH) */
+    }
+    Veth veth = 9;
```
ondrej-fabry: Fields `veth` and `tap` are also candidates for `oneof` field type..

plugins/linuxv2/model/l3/l3.proto
```diff
@@ -0,0 +1,27 @@
+syntax = "proto3";
+
+package l3;
+
+/* static ip route */
```
ondrej-fabry: These comments before message are not helpful at all, you can remove them.

plugins/linuxv2/model/l3/l3.proto
```diff
@@ -0,0 +1,27 @@
...
+    string outgoing_interface = 1; /* outgoing interface logical name (mandatory) */
+
+    enum Scope {
+        GLOBAL = 0;
+        SITE = 1;
+        LINK = 2;
+        HOST = 3;
+    }
+    Scope scope = 2;               /* the scope of the area where the link is valid */
+
```
ondrej-fabry: I don't know if it happened by accident, but it seems you missed some of the fields for static routes that seemed relevant, like: `src_ip_addr` `table` (VRF) and `namespace`.

plugins/linuxv2/model/l3/l3.proto
```diff
@@ -0,0 +1,27 @@
...
+    string dst_network = 3;        /* destination network address in the format <address>/<prefix> (mandatory) */
+    string gw_addr = 4;            /* gateway IP address */
+    uint32 metric = 5;             /* routing metric (weight) */
+}
+
+/* static arp entry */
+message LinuxStaticARPEntry {
+    string interface = 1;          /* logical name of the interface to which this neighbour is attached (mandatory) */
+    string ip_addr = 2;            /* ip address (mandatory) */
+    string hw_address = 3;         /* mac address (mandatory) */
```
ondrej-fabry: Field `namespace` missing here.

plugins/linuxv2/model/namespace/namespace.proto
milanlenco: OK, but will also add our own NETNSA_MICROSERVICE

plugins/linuxv2/model/interfaces/interfaces.proto
milanlenco: Nope, you can have interfaces with the same host name, but in different namespaces - like in K8s all pods must have eth0 interface.

plugins/linuxv2/model/l3/l3.proto
milanlenco: Namespace is given by the interface, not needed to mention that again.

plugins/linuxv2/model/namespace/namespace.proto
ondrej-fabry: Isn't that what I suggested?

clientv2/linux/data_change_api.go
```diff
@@ -0,0 +1,170 @@
+// Copyright (c) 2017 Cisco and/or its affiliates.
```
VladoLavor: maybe some code is just copied from the clientv1, but i would put current year to the licence header, everywhere

plugins/linuxv2/ifplugin/linuxcalls/admin_linuxcalls.go
```diff
@@ -0,0 +1,63 @@
...
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+// +build !windows,!darwin
+
+// Copyright (c) 2017 Cisco and/or its affiliates.
```
VladoLavor: why there are two headers?

plugins/linuxv2/l3plugin/linuxcalls/arp_linuxcalls.go
```diff
@@ -0,0 +1,52 @@
...
+package linuxcalls
+
+import (
+	"time"
+
+	"github.com/vishvananda/netlink"
+)
+
+// SetARPEntry updates existing arp entry
+func (handler *NetLinkHandler) SetARPEntry(arpEntry *netlink.Neigh) error {
```
VladoLavor: same for handlers, let's call the receiver `h`

plugins/linuxv2/model/l3/l3.proto
```diff
@@ -0,0 +1,27 @@
+syntax = "proto3";
+
+package l3;
+
+/* static ip route */
+message LinuxStaticRoute {
```
VladoLavor: question - I miss name here, will it be generated from other fields? Same for arp.

plugins/linuxv2/model/namespace/namespace.proto
milanlenco: yep, sorry, didn't read it right :)

plugins/linuxv2/ifplugin/linuxcalls/admin_linuxcalls.go
milanlenco: This file was just copied from the original linux plugin - somebody must have put it there by mistake.

plugins/linuxv2/model/l3/l3.proto
ondrej-fabry: @milanlenco mentioned in our last discussion that he chose to remove the name field for things like routes, because we don't set it to southbound in any way and the name does not actually represent the route and is not included in key either

plugins/linuxv2/ifplugin/linuxcalls/admin_linuxcalls.go
ondrej-fabry: You should rebase on latest pantheon-dev probably.. I think this was fixed by @VladoLavor already

plugins/linuxv2/model/namespace/namespace.proto
ondrej-fabry: Just to clarify.. I did not mean to suggest to use `NETNSA_` prefix. My intention was to follow the same order, basically have same value for same type.

plugins/linuxv2/model/l3/l3.proto
milanlenco: For ARPs and routes I chosed what we do in VPP l3 plugin - not naming things, but identifying them by their primary fields. This is for several reasons:
 1. dumping is simplified - you can build value + key from the dumped data without needing support for metadata (like tags, which in Linux are not available)
2. arps and routes may be created externally (e.g. dynamic entries), but the same route/arp by value will have the same ID, regardless of whether it was configured by the agent or came from somewhere else - thus the scheduler will be able to compare and match the same items (thus not trying to configure the same twice)
3. With name as the only primary key, you can create two entries which are effectively the same (thus ending with add-error for the second one) - I wanted to prevent that from happening
4. In many cases you will end up building the name from the primary fields anyway

I would say that configuration items should be logically named only when:
1. there can be multiple items of the same value (i.e. no primary keys in the value) - e.g. ACLs
2. there are too many fields needed to include in the key for unique identification (and tagging is supported)
2. there is more to the item than just it's value - e.g. interface "goes" somewhere
3. the item is already "logical" - e.g. bridge domain
etc.

milanlenco:
The review comments were reflected into the PR.
Also, the linux plugin was rebased on top of the "revisited" KVScheduler (https://github.com/ligato/cn-infra/pull/343).

plugins/linuxv2/ifplugin/descriptor/interface_veth.go
```diff
@@ -0,0 +1,194 @@
...
+	vethName = aliasParts[0]
+	if len(aliasParts) > 0 {
+		peerName = aliasParts[1]
+	}
+	return
+}
+
+// getVethPeerName returns the name of the peer interface from the configuration.
+func getVethPeerName(linuxIf *interfaces.LinuxInterface) string {
+	link, ok := linuxIf.Link.(*interfaces.LinuxInterface_Veth)
```
ondrej-fabry: This is completely unnecessary. Oneof field provides methods for all of this. You can simply just call `linuxIf.GetVeth().GetPeerName()` and it will do the same thing that this function does.

PR #900, opened 2018-09-24
Support for building of docker images for ARM64 platform

stanislav-chlebec:
The changes are inspired by this pull request 
https://github.com/contiv/vpp/commit/535b0614eec1ba8cd0b743d040ead7742c43b14c

There are covered both docker/dev and docker/prod images.

The scripts were rewritten with effort to preserve the previous behaviour when it is possible to pre-set some environment variables then export them and use the newly defined values in scripts instead of defined default values.

Warning: there was removed the sudo command while calling docker command in docker/prod/build.sh what can make problem on not sufficiently set up machines...

### DIFF
--- a/docker/dev/build.sh
+++ b/docker/dev/build.sh
@@ -1,4 +1,7 @@
 #!/bin/bash
+# Before run of this script you can set environmental variables
+# IMAGE_TAG, DOCKERFILE, BASE_IMG, GOLANG_OS_ARCH, .. then  export them
+# and to use defined values instead of default ones
 
 cd "$(dirname "$0")"
 
@@ -17,17 +20,22 @@ BASE_IMG=${BASE_IMG:-'ubuntu:18.04'}
 
 BUILDARCH=`uname -m`
 
-if [ ${BUILDARCH} = "aarch64" ] ; then
-  IMAGE_TAG=${IMAGE_TAG:-'dev_vpp_agent-arm64'}
-  GOLANG_OS_ARCH=${GOLANG_OS_ARCH:-'linux-arm64'}
-fi
+case "$BUILDARCH" in
+  "aarch64" )
+    IMAGE_TAG=${IMAGE_TAG:-'dev_vpp_agent-arm64'}
+    GOLANG_OS_ARCH=${GOLANG_OS_ARCH:-'linux-arm64'}
+    ;;
 
-
-if [ ${BUILDARCH} = "x86_64" ] ; then
-  # for AMD64 platform is used the default image (without suffix -amd64)
-  IMAGE_TAG=${IMAGE_TAG:-'dev_vpp_agent'}
-  GOLANG_OS_ARCH=${GOLANG_OS_ARCH:-'linux-amd64'}
-fi
+  "x86_64" )
+    # for AMD64 platform is used the default image (without suffix -amd64)
+    IMAGE_TAG=${IMAGE_TAG:-'dev_vpp_agent'}
+    GOLANG_OS_ARCH=${GOLANG_OS_ARCH:-'linux-amd64'}
+    ;;
+  * )
+    echo "Architecture ${BUILDARCH} is not supported."
+    exit
+    ;;
+esac
 
 
 source ../../vpp.env
@@ -61,6 +69,6 @@ docker build -f ${DOCKERFILE} \
     ${DOCKER_BUILD_ARGS} ../..
 
 if [[ ${BUILDARCH} = "x86_64" && ${IMAGE_TAG} = "dev_vpp_agent" ]] ; then
-  # tag also explicit docker image on AMD64 platform
+  # create docker image tagged with -amd64 suffix for AMD64 platform
   docker tag  ${IMAGE_TAG}:latest dev_vpp_agent-amd64:latest
 fi
--- a/docker/dev/build.sh
+++ b/docker/dev/build.sh
@@ -4,11 +4,31 @@ cd "$(dirname "$0")"
 
 set -e
 
-IMAGE_TAG=${IMAGE_TAG:-'dev_vpp_agent'}
+#IMAGE_TAG=${IMAGE_TAG:-'dev_vpp_agent'}
 DOCKERFILE=${DOCKERFILE:-'Dockerfile'}
 
 BASE_IMG=${BASE_IMG:-'ubuntu:18.04'}
-GOLANG_OS_ARCH=${GOLANG_OS_ARCH:-'linux-amd64'}
+#GOLANG_OS_ARCH=${GOLANG_OS_ARCH:-'linux-amd64'}
+
+#To prepare for future fat manifest image by multi-arch manifest,
+#now build the docker image with its arch
+#For fat manifest, please refer
+#https://docs.docker.com/registry/spec/manifest-v2-2/#example-manifest-list
+
+BUILDARCH=`uname -m`
+
+if [ ${BUILDARCH} = "aarch64" ] ; then
+  IMAGE_TAG=${IMAGE_TAG:-'dev_vpp_agent-arm64'}
+  GOLANG_OS_ARCH=${GOLANG_OS_ARCH:-'linux-arm64'}
+fi
+
+
+if [ ${BUILDARCH} = "x86_64" ] ; then
+  # for AMD64 platform is used the default image (without suffix -amd64)
+  IMAGE_TAG=${IMAGE_TAG:-'dev_vpp_agent'}
+  GOLANG_OS_ARCH=${GOLANG_OS_ARCH:-'linux-amd64'}
+fi
+
 
 source ../../vpp.env
 VPP_DEBUG_DEB=${VPP_DEBUG_DEB:-}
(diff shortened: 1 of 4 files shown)


docker/dev/build.sh
```diff
@@ -37,3 +67,23 @@ docker build -f ${DOCKERFILE} \
     --build-arg VERSION=${VERSION} \
     --build-arg COMMIT=${COMMIT} \
     ${DOCKER_BUILD_ARGS} ../..
+
+echo "To push to repository please use command:"
```
ondrej-fabry: I would prefer not to put info about tag/pushing in build script for the image. That's needed only needed in deployment which we do quite differently.

docker/dev/build.sh
```diff
@@ -1,14 +1,42 @@
...
+# Before run of this script you can set environmental variables
+# IMAGE_TAG, DOCKERFILE, BASE_IMG, GOLANG_OS_ARCH, .. then  export them
+# and to use defined values instead of default ones
 
 cd "$(dirname "$0")"
 
 set -e
 
-IMAGE_TAG=${IMAGE_TAG:-'dev_vpp_agent'}
+#IMAGE_TAG=${IMAGE_TAG:-'dev_vpp_agent'}
```
ondrej-fabry: Please keep the image-tag name, otherwise this is gonna break build for production image, the name is used as base image there. Let's keep this simple and just use `dev_vpp_agent`, it's local name anyway.

docker/dev/build.sh
```diff
@@ -1,14 +1,42 @@
 #!/bin/bash
+# Before run of this script you can set environmental variables
```
ondrej-fabry: Actually some prefer not run it this way and only set these variables for single command execution, like this:

```sh
$ BASE_IMG=ubuntu:18.04 ./build.sh
```

docker/prod/build.sh
```diff
@@ -1,9 +1,59 @@
...
+#https://docs.docker.com/registry/spec/manifest-v2-2/#example-manifest-list
 
-sudo docker build  ${DOCKER_BUILD_ARGS} --tag ${IMAGE_TAG} .
+BUILDARCH=`uname -m`
+
+case "$BUILDARCH" in
+  "aarch64" )
+    IMAGE_TAG=${IMAGE_TAG:-'prod_vpp_agent-arm64'}
+    # Dockerfile  for prod_vpp_agent expects that dev-vpp-agent:latest image is built
+    docker tag dev_vpp_agent-arm64:latest dev_vpp_agent:latest
```
ondrej-fabry: Now I see you are retagging here, but this just unnecessarily complicates overall build process. Always try to minimize parts that don't really help, this just make script longer with extra commands to execute and actually does nothing useful. Keep local image name just `dev_vpp_agent` and `prod_vpp_agent`.

docker/dev/build.sh
stanislav-chlebec: Well I remove that. I will created the push_image.sh instead which directly does pushing.

docker/dev/build.sh
stanislav-chlebec: ok

docker/prod/build.sh
stanislav-chlebec: ok

docker/dev/build.sh
stanislav-chlebec: ok

docker/dev/push_image.sh
```diff
@@ -0,0 +1,77 @@
...
+    ;;
+
+  "x86_64" )
+    # for AMD64 platform is also used the default image (without suffix -amd64)
+    DEFAULT_IMAGE_NAME='dev-vpp-agent'
+    IMAGE_NAME='dev-vpp-agent-amd64'
+    ;;
+  * )
+    echo "Architecture ${BUILDARCH} is not supported."
+    exit
```
ondrej-fabry: This should most likely exit with 1 so it is obvious that there is some problem.

docker/dev/push_image.sh
```diff
@@ -0,0 +1,77 @@
...
+
+case "$BUILDARCH" in
+  "aarch64" )
+    IMAGE_NAME='dev-vpp-agent-arm64'
+    ;;
+
+  "x86_64" )
+    # for AMD64 platform is also used the default image (without suffix -amd64)
+    DEFAULT_IMAGE_NAME='dev-vpp-agent'
+    IMAGE_NAME='dev-vpp-agent-amd64'
```
ondrej-fabry: You could only store the suffix here and reuse `DEFAULT_IMAGE_NAME`.

docker/dev/push_image.sh
```diff
@@ -0,0 +1,77 @@
...
+BUILDARCH=`uname -m`
+
+case "$BUILDARCH" in
+  "aarch64" )
+    IMAGE_NAME='dev-vpp-agent-arm64'
+    ;;
+
+  "x86_64" )
+    # for AMD64 platform is also used the default image (without suffix -amd64)
+    DEFAULT_IMAGE_NAME='dev-vpp-agent'
```
ondrej-fabry: This should most likely be defined just after `REPO_OWNER`.

docker/dev/push_image.sh
```diff
@@ -0,0 +1,77 @@
...
+echo "=============================="
+echo "Architecture: ${BUILDARCH}"
+echo "=============================="
+
+case "${BRANCH_NAME}" in
+  "master" )
+    if [ ${BRANCH_HEAD_TAG} != "undefined" ] ; then
+      if [ ${BUILDARCH} = "x86_64" ] ; then
+        # for AMD64 platform is used also the default image (without suffix -amd64)
+        docker tag ${LOCAL_IMAGE}:latest ${REPO_OWNER}/${DEFAULT_IMAGE_NAME}:${BRANCH_HEAD_TAG}
```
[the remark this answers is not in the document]
stanislav-chlebec: Done.

docker/dev/push_image.sh
stanislav-chlebec: Done.

docker/dev/push_image.sh
stanislav-chlebec: Done.

docker/dev/push_image.sh
stanislav-chlebec: Done.

docker/dev/push_image.sh
```diff
@@ -0,0 +1,78 @@
...
+
+# detect branch name
+BRANCH_NAME="$(git symbolic-ref HEAD 2>/dev/null)" || BRANCH_NAME="(unnamed branch)"     # detached HEAD
+BRANCH_NAME=${BRANCH_NAME##refs/heads/}
+BRANCH_HEAD_TAG=${BRANCH_HEAD_TAG:-"`git name-rev --name-only --tags HEAD`"}
+VERSION=$(git describe --always --tags --dirty)
+
+LOCAL_IMAGE=${LOCAL_IMAGE:-'dev_vpp_agent:latest'}
+REPO_OWNER=${REPO_OWNER:-'ligato'}
+DEFAULT_IMAGE_NAME=${DEFAULT_IMAGE_NAME:-'dev-vpp-agent'}
```
ondrej-fabry: And as a final thing, just shorten this to `IMAGE_NAME` since we don't have that variable now and I'll merge it.

stanislav-chlebec: Done:-)
Now it is almost perfect.

docker/prod/push_image.sh
```diff
@@ -0,0 +1,11 @@
...
+# Usage: examples
+#    ./push_image.sh
+#    BRANCH_HEAD_TAG='git describe' ./push_image.sh
+#    REPO_OWNER=stanislavchlebec BRANCH_HEAD_TAG=`git describe` ./push_image.sh
+#    LOCAL_IMAGE='prod_vpp_agent:latest' DEFAULT_IMAGE_NAME='vpp-agent' ./push_image.sh
+
+# Warning: use only IMMEDIATELY after docker/dev/build.sh to prevent INCONSISTENCIES such as 
+#          a) after building image you switch to other branch which will result in mismatch of version of image and its tag
+#          b) you do not build the new image but only simply run this script which will result in mismatch version of image and its tag because the image is older than repository 
+LOCAL_IMAGE='prod_vpp_agent:latest' DEFAULT_IMAGE_NAME='vpp-agent' ../dev/push_image.sh
```
ondrej-fabry: Forgot to change here. Perhaps the push script should be only one and located in `docker` directory directly.

stanislav-chlebec: I would prefer to leave it like it is.
The user then does not have to write variables before script
so IMO it has sense to make it like it is.

stanislav-chlebec: The next thing is that how it is described in the beginning of the script there is close connection to the build.sh script - they IMHO should be one next to other one.

ondrej-fabry: Okay then. `DEFAULT_IMAGE_NAME` needs to be fixed anyway though.

stanislav-chlebec: Yes - that is what you refered that I forgot:-) I did not realised.
I do it.

PR #921, opened 2018-10-02
VPP interface plugin v.2 + latest pantheon-dev

milanlenco:
TODO:
- [x]  Modify, Dump
- [x]  DHCP
- [x]  interface state data  (not using scheduler, mostly just copy&paste)
- [x] Prepare scheduler for error publishing
- [x] testing

Interface errors will be published using a generic plugin - not just for interfaces and BDs (watching errors from scheduler).

I have also renamed AUTO_TAP to TAP_TO_VPP to make it clearer what it is (it cannot be any TAP, but specifically for connecting with VPP) - also the reference to VPP-side of the TAP is now via the logical name.


plugins/vppv2/ifplugin/descriptor/dhcp.go
```diff
@@ -202,9 +217,9 @@ func (d *DHCPDescriptor) Dump(correlate []scheduler.KVWithMetadata) ([]scheduler
 }
 
 // watchDHCPNotifications watches and processes DHCP notifications.
-func (d *DHCPDescriptor) watchDHCPNotifications(ctx context.Context, wg sync.WaitGroup, dhcpChan chan govppapi.Message) {
-	wg.Add(1)
-	defer wg.Done()
+func (d *DHCPDescriptor) watchDHCPNotifications(ctx context.Context, dhcpChan chan govppapi.Message) {
+	d.wg.Add(1)
```
ondrej-fabry: This creates race condition. `Add` is supposed to be called before goroutine is started.

Check out docs and example for more info: https://godoc.org/sync#example-WaitGroup

PR #943, opened 2018-10-10
L2 Plugin v.2

milanlenco:
- [x] model
- [x] vpp calls
- [x] plugin skeleton
- [x] BD descriptor
- [x] FIB descriptor
- [x] xConnect descriptor
- [x] example
- [x] testing


plugins/vppv2/l2plugin/descriptor/bd_interface.go
```diff
@@ -0,0 +1,135 @@
...
+	"github.com/ligato/vpp-agent/plugins/vppv2/l2plugin/descriptor/adapter"
+	"github.com/ligato/vpp-agent/plugins/vppv2/l2plugin/vppcalls"
+	"github.com/ligato/vpp-agent/plugins/vppv2/model/l2"
+	"github.com/ligato/vpp-agent/plugins/vppv2/model/interfaces"
+)
+
+const (
+	// BDInterfaceDescriptorName is the name of the descriptor for bindings between
+	// VPP bridge domains and interfaces.
+	BDInterfaceDescriptorName = "vpp-bridge-domain"
```
ondrej-fabry: Shouldn't this be `vpp-bridge-domain-interface`?

plugins/vppv2/l2plugin/descriptor/bridgedomain.go
```diff
@@ -0,0 +1,344 @@
...
+	}
+
+	// ARP termination entries
+	obsoleteARPs, newARPs := calculateARPDiff(oldBD.GetArpTerminationTable(), newBD.GetArpTerminationTable())
+	return len(obsoleteARPs) == 0 && len(newARPs) == 0
+}
+
+// MetadataFactory is a factory for index-map customized for VPP bridge domains.
+func (d *BridgeDomainDescriptor) MetadataFactory() idxmap.NamedMappingRW {
+	return idxvpp2.NewNameToIndex(logrus.DefaultLogger(), "vpp-bd-index", nil)
```
ondrej-fabry: This should probably use `d.log` instead of the default logger.

plugins/vppv2/model/l2/xconnect.proto
```diff
@@ -0,0 +1,8 @@
+syntax = "proto3";
+
+package l2;
+
+message XConnectPair {
+    string receive_interface = 2;               /* receive interface name */
```
ondrej-fabry: Should start from 1.

plugins/vppv2/l2plugin/descriptor/bd_interface.go
```diff
@@ -0,0 +1,135 @@
...
+		d.log.Error(err)
+		return err
+
+	}
+	return nil
+}
+
+// ModifyWithRecreate returns always true - a change in BVI or SHG is always performed
+// via Delete+Add.
+func (d *BDInterfaceDescriptor) ModifyWithRecreate(key string, oldBDIface, newBDIface *l2.BridgeDomain_Interface, metadata interface{}) bool {
```
ondrej-fabry: Could this be omitted, left nil in descriptor and then scheduler could figure it out automatically when Modify is nil as well.

milanlenco: I guess it could, this way it is just explicit that Modify wasn't excluded by mistake,

PR #988, opened 2018-11-06
Fixes for bugs in v2 plugins found with Contiv

milanlenco:
With these fixes, the resync for Contiv is almost "empty" (when it should be), the only operation is modify on IPNeigh:

  * planned operations:
      1. MODIFY:
          - key: vpp/config/v2/ipneigh
          - prev-value: { mode:IPv4 scan_interval:1 max_proc_time:20 max_update:10 scan_int_delay:1 stale_threshold:4  } 
          - new-value: { mode:IPv4 scan_interval:1 stale_threshold:4  }

Default values need to be considered in the comparison function, Ondrej would you take care of it?

### DIFF
--- a/plugins/vppv2/aclplugin/descriptor/acl.go
+++ b/plugins/vppv2/aclplugin/descriptor/acl.go
@@ -15,6 +15,8 @@
 package descriptor
 
 import (
+	"bytes"
+	"net"
 	"strings"
 
 	"github.com/gogo/protobuf/proto"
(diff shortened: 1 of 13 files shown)


plugins/vppv2/natplugin/vppcalls/nat_vppcalls.go
```diff
@@ -36,6 +38,10 @@ const (
 	NoInterface = ^uint32(0)
 	// Maximal length of tag
 	maxTagLen = 64
+	// label added to tags of DNAT mappings with external IP from the pool.
+	extIPFromPoolLabel = "POOL" // keep it short, there is a limit on the maximum tag length
```
ondrej-fabry: So it seems we can't work without adding some metadata to the tag. Could you please explain to me what this represents? (I have limited experience with NAT)

Perhaps it would be better keep only `P` instead of `POOL` to save 3 characters.

plugins/vppv2/l3plugin/descriptor/static_route.go
```diff
@@ -89,7 +92,7 @@ func (d *RouteDescriptor) EquivalentRoutes(key string, oldRoute, newRoute *l3.St
 		oldRoute.GetVrfId() != newRoute.GetVrfId() ||
 		oldRoute.GetViaVrfId() != newRoute.GetViaVrfId() ||
 		oldRoute.GetOutgoingInterface() != newRoute.GetOutgoingInterface() ||
-		oldRoute.GetWeight() != newRoute.GetWeight() ||
+		getWeight(oldRoute) != getWeight(newRoute) ||
```
ondrej-fabry: Why doesn't this compare default value as `0` and then in vppcalls package set `1` in the binapi request instead of `0`.

plugins/vppv2/aclplugin/vppcalls/interfaces_vppcalls.go
```diff
@@ -192,6 +207,9 @@ func (h *ACLVppHandler) DeleteMACIPACLFromInterface(aclIndex uint32, ifName stri
 	if err != nil {
 		return fmt.Errorf("failed to delete interface %d from MACIP ACL (L2) %d: %v", ifIdx, aclIndex, err)
 	}
+	if reply.Retval != 0 {
```
ondrej-fabry: These were excluded intentionally. The GoVPP's `ReceiveReply` will return error if it's non-nil. Please remove it and I will cleanup all the other vppcalls to remove it there as well.

milanlenco: OK, will do

plugins/vppv2/l3plugin/descriptor/static_route.go
milanlenco: No this fixes comparison between what is dumped and what is defined by NB,
The unspecified fields may remain empty in the binapi calls, but dump will return them set to the default values and resync would trigger modify if we do not consider them in the comparison function.

plugins/vppv2/natplugin/vppcalls/nat_vppcalls.go
milanlenco: This "ExternalIpFromPool" thing is actually a temporary workaround for bug in the VPP NAT plugin - Matus will soon provide a fix. The plan is to revert this before we merge dev into master, but for now I needed it to unblock Contiv refactoring.
Otherwise we really do not need any extra labels to add into the tags.

plugins/vppv2/aclplugin/descriptor/acl.go
```diff
@@ -99,14 +101,68 @@ func (d *ACLDescriptor) EquivalentACLs(key string, oldACL, newACL *acl.Acl) bool
...
+// equivalentACLRules compares two ACL rules, handling the cases of unspecified
+// source/destination networks.
+func (d *ACLDescriptor) equivalentACLRules(rule1, rule2 *acl.Acl_Rule) bool {
+	if rule1.Action != rule2.Action || !proto.Equal(rule1.MacipRule, rule2.MacipRule) {
+		return false
+	}
+
+	// IP Rule
+	if rule1.IpRule == nil || rule2.IpRule == nil {
+		return rule1.IpRule == rule2.IpRule
```
ondrej-fabry: This compares pointers only. You probably wanted to use `proto.Equal`?

plugins/vppv2/aclplugin/descriptor/acl.go
```diff
@@ -99,14 +101,68 @@ func (d *ACLDescriptor) EquivalentACLs(key string, oldACL, newACL *acl.Acl) bool
...
 		}
 	}
 
 	return true
 }
 
+// equivalentACLRules compares two ACL rules, handling the cases of unspecified
+// source/destination networks.
+func (d *ACLDescriptor) equivalentACLRules(rule1, rule2 *acl.Acl_Rule) bool {
+	if rule1.Action != rule2.Action || !proto.Equal(rule1.MacipRule, rule2.MacipRule) {
```
ondrej-fabry: For better code clarity I would suggest to split checks like this which don't have much in common and rather group them by specific sections/fields.

plugins/vppv2/aclplugin/descriptor/acl.go
```diff
@@ -99,14 +101,68 @@ func (d *ACLDescriptor) EquivalentACLs(key string, oldACL, newACL *acl.Acl) bool
...
+		return false
+	}
+
+	// IP Rule
+	if rule1.IpRule == nil || rule2.IpRule == nil {
```
ondrej-fabry: Another trivial one, you could use getters in the code below (e.g. `rule1.GetIpRule().GetIcmp()`) to eliminate this check.

PR #996, opened 2018-11-06
Cleanup v2 code and run with KVScheduler by default

ondrej-fabry:


plugins/vppv2/ifplugin/descriptor/interface.go
```diff
@@ -120,7 +119,7 @@ type InterfaceDescriptor struct {
 	memifSocketToID        map[string]uint32 // memif socket filename to ID map (all known sockets)
 	defaultMemifSocketPath string
 	ethernetIfs            map[string]uint32 // name-to-index map of ethernet interfaces
-	                                         // (entry is not removed even if interface is un-configured)
```
milanlenco: Probably auto-format, would you put this comment onto the same line as ethernetIfs, so that it is clear what it belongs to.

PR #997, opened 2018-11-07
Do not leave file opened on unsuccessful ns switch

lukasmacko:


plugins/linux/nsplugin/ns_handler.go
```diff
@@ -182,6 +182,7 @@ func (h *NsHandler) SwitchNamespace(ns *Namespace, ctx *NamespaceMgmtCtx) (rever
 	// Get network namespace file descriptor.
 	nsHandle, err = h.GetOrCreateNamespace(ns)
 	if err != nil {
+		origns.Close()
```
ondrej-fabry: Is this the cause of the issue that was occurring in the tests so often?

rastislavszabo: Yes it was. We may want to cherry-pick this into the master and vpp/1810 branch as well.

ondrej-fabry: @VladoLavor could you please take care of this?

PR #1022, opened 2018-11-20
Introduce Model (proto+key) with new generic NB API

ondrej-fabry:


api/api.proto
```diff
@@ -0,0 +1,73 @@
...
+
+message ModelSpec {
+    string module = 1;
+    string class = 2;
+    string version = 3;
+    string kind = 4;
+}
+
+message Model {
+    google.protobuf.Any any = 1;
```
edwarnicke: This is slick, but what is the client experience like using Any, particularly in Golang?

api/api.proto
```diff
@@ -0,0 +1,73 @@
...
+
+message SyncItem {
+    Model model = 1;
+    bool delete = 2;
+}
+
+message SyncItemResult {
+    string id = 1;
+    string status = 2;
+    string error = 3;
```
edwarnicke: Why not simply rely on the normal grpc error mechanism?  Why a separate string here?

api/api.proto
```diff
@@ -0,0 +1,73 @@
...
+}
+
+message Model {
+    google.protobuf.Any any = 1;
+    map<string,string> metadata = 2;
+}
+
+service SyncService {
+    rpc Sync (SyncRequest) returns (SyncResponse) {};
+    rpc Obtain (ObtainRequest) returns (ObtainResponse) {};
```
edwarnicke: Semantically, what is 'Obtain' and what is 'Sync' ?  Most devs don't think in quite these terms, so might be useful to put a very fine point on what we mean by them.

api/models/linux/namespace/namespace.proto
```diff
@@ -0,0 +1,27 @@
...
+
+import "github.com/gogo/protobuf/gogoproto/gogo.proto";
+option (gogoproto.messagename_all) = true;
+
+message NetNamespace {
+    enum ReferenceType {
+        UNDEFINED = 0;
+        NSID = 1;         /* named namespace */
+        PID = 2;          /* namespace of a given process */
+        FD = 3;           /* namespace referenced by a file handle */
```
edwarnicke: Is this really an FD?  Or is it a filename?

api/models/linux/namespace/namespace.proto
```diff
@@ -0,0 +1,27 @@
...
+
+option go_package = "github.com/ligato/vpp-agent/api/models/linux/namespace;linux_namespace";
+
+import "github.com/gogo/protobuf/gogoproto/gogo.proto";
+option (gogoproto.messagename_all) = true;
+
+message NetNamespace {
+    enum ReferenceType {
+        UNDEFINED = 0;
+        NSID = 1;         /* named namespace */
```
edwarnicke: When using NSID... how are you figuring out which directory to look for the NSID file for the named namespace?

api/models/vpp/interfaces/state.proto
```diff
@@ -1,8 +1,13 @@
...
 
-import "interface.proto";
+option go_package = "github.com/ligato/vpp-agent/api/models/vpp/interfaces;vpp_interfaces";
+
+import "github.com/gogo/protobuf/gogoproto/gogo.proto";
+option (gogoproto.messagename_all) = true;
+
+import "vpp/interfaces/interface.proto";
 
 message InterfaceState {
```
edwarnicke: Curious, why a string and not an enum?

edwarnicke:
So what would relatively simple things look like for the user?  For example, if I want to create a Tap interface into a namespace with a host side IP.  What message do I send for that?  Is it one thing or two?

api/api.proto
ondrej-fabry: This is meant to define status/error for the specific model item that was synced. The normal grpc error mechanism will be used for the overall result of the request.

api/api.proto
ondrej-fabry: It's very simple. The `models` package also provides short-hands to make it single call for marshaling/unmarshalling all models: https://github.com/ligato/vpp-agent/blob/0fbdf09e631368e282e2baffd365a212157bc89f/api/models/encoding.go#L29-L68

api/models/vpp/interfaces/state.proto
ondrej-fabry: Not sure what you are referring to here. This `InterfaceState` is remnant from the old behavior and might be deprecated to provide uniform way for the state of any configuration.

ondrej-fabry:
@edwarnicke For example with remote configuration via gRPC, see: https://github.com/ligato/vpp-agent/blob/0fbdf09e631368e282e2baffd365a212157bc89f/examples/grpc_vpp/remote_client/main.go#L151-L165

Anyway, this PR is mostly about defining generic NB API and clients to provide single definition for all types of configuration currently supported and also with any extended configuration (e.g. custom VPP plugin) or even configuration of new SB beside vpp/linux (e.g. strongswan).

api/api.proto
edwarnicke: @ondrej-fabry That's super painful actually.

You expect to be able to just use types in go when using things like GRPC... Short-circuiting that by requiring unnatural additional Unmarshal calls is kind of awful.

Among other things, it requires any recipient of an object containing Models to understand its an opaque proto thing.  That's a big usability regression on the current API, which gives you nice usable Go objects at all levels that you can just pass around and use.

api/api.proto
edwarnicke: Ah... that makes sense

api/api.proto
edwarnicke: So Sync is 'Set' and 'Obtain' is 'Get' ?  Those might be clearer :)

api/models/vpp/interfaces/state.proto
edwarnicke: "History" is always a reason :)  I was mostly curious as to the reason, not arguing one way or the other.  I've seen State handled both ways :)

edwarnicke:
> @edwarnicke For example with remote configuration via gRPC, see:
> 
> [vpp-agent/examples/grpc_vpp/remote_client/main.go](https://github.com/ligato/vpp-agent/blob/0fbdf09e631368e282e2baffd365a212157bc89f/examples/grpc_vpp/remote_client/main.go#L151-L165)
> 
> Lines 151 to 165 in [0fbdf09](/ligato/vpp-agent/commit/0fbdf09e631368e282e2baffd365a212157bc89f)
> 
>  c := remoteclient.NewClientGRPC(api.NewSyncServiceClient(conn)) 
>   
>  req := c.ResyncRequest() 
>  req.Put(memif1, memif2) 
>  if err := req.Send(context.Background()); err != nil { 
>  	log.Fatalln(err) 
>  } 
>   
>  time.Sleep(time.Second * 5) 
>   
>  req2 := c.ChangeRequest() 
>  req2.Delete(memif1) 
>  if err := req2.Send(context.Background()); err != nil { 
>  	log.Fatalln(err) 
>  } 
> Anyway, this PR is mostly about defining generic NB API and clients to provide single definition for all types of configuration currently supported and also with any extended configuration (e.g. custom VPP plugin) or even configuration of new SB beside vpp/linux (e.g. strongswan).

What does it look like to put memif1 and memif2 together?
Do I still have to do weird correlations for tap interfaces?
etc

api/api.proto
ondrej-fabry: To me, GET/SET is used for things which are equal for both operations, meaning that via **GET** you get the exact value the was set via **SET**. Which is not the case here, so I tried to use different wording. However I have to agree that I'm not very fond of this exact wording to be honest.

Perhaps, **Configure**/**Retrieve** are more suitable candidates here.

ondrej-fabry:
@edwarnicke What you are describing is kind of functionality that can be provided by some higher level plugin. Or perhaps some new `Connection` model which could serve for various different combinations and their distinct implementations.

Until then, the VPP/Linux plugins try to mirror the southband to provide the most transparent access to it's existing capabilities.

api/api.proto
edwarnicke: Or perhaps just Read/Write?

plugins/kvscheduler/refresh.go
```diff
@@ -97,8 +97,15 @@ func (scheduler *Scheduler) refreshGraph(graphW graph.RWAccess, keys utils.KeySe
 			if len(dump) == 1 {
 				plural = ""
 			}
-			scheduler.Log.Debugf("Descriptor %s dumped %d value%s: %v",
-				descriptor.Name, len(dump), plural, dump)
+
+			var dumpList string
+			for _, d := range dump {
+				dumpList += fmt.Sprintf("\n - %+v", d)
```
milanlenco: hmm, I didn't include the new line because the logger ignores the character and puts everything into one line - at least that is the case in Contiv - or can the behavior be changed in the logger configuration?

ondrej-fabry: Yes, the logger output can be changed via formatters. 

And this way the formatter used for console output is properly formatted to make it easier to read during development while it doesn't alter the output that is currently printed for non-console (pipe to file) very much beside replacing newlines with `\n` character.

api/dataconfigurator/dataconfigurator.proto
```diff
@@ -0,0 +1,39 @@
...
+import "models/vpp/vpp.proto";
+import "models/linux/linux.proto";
+
+message ConfigData {
+    vpp.Config vpp = 1;
+    linux.Config linux = 2;
+}
+
+service DataConfigurator {
+    rpc Resync(ConfigData) returns (ResyncResponse);
```
edwarnicke: Don't take this comment to seriously... but if you:

Resync -> Create + Delete (${some value for ConfigData that deletes *all* data)
Get -> Read

You are left with the super familiar 'CRUD' ...

api/models/linux/linux.proto
```diff
@@ -0,0 +1,18 @@
+syntax = "proto3";
+
+package linux;
+
+option go_package = "github.com/ligato/vpp-agent/api/models/linux;linux";
+
+import "github.com/gogo/protobuf/gogoproto/gogo.proto";
+option (gogoproto.messagename_all) = true;
+
+import "models/linux/interfaces/interface.proto";
```
edwarnicke: This is so much better than all the embedded messages previously!

api/models/linux/namespace/namespace.proto
edwarnicke: ^^^^^^^

api/models/linux/namespace/namespace.proto
edwarnicke: ^^^^^^

api/dataconfigurator/dataconfigurator.proto
ondrej-fabry: Yeah. Plan is to hide _resync_ from outside, since it's really not common way to to stuff.

edwarnicke: All good :)

api/models/linux/namespace/namespace.proto
milanlenco: This is done by the external package `vishvananda/netns`, which has hard-coded `/var/run/netns` directory:
https://github.com/ligato/vpp-agent/blob/dev/vendor/github.com/vishvananda/netns/netns_linux.go#L80

api/models/linux/namespace/namespace.proto
milanlenco: This is actually file path, so perhaps `FILENAME` would be better?

client/client_api.go
```diff
@@ -0,0 +1,30 @@
...
+
+// ConfigClient defines the client-side interface for config service.
+type ConfigClient interface {
+	// ActiveModels retrieves list of active modules.
+	ActiveModels() (map[string][]api.ModelInfo, error)
+
+	SetConfig(resync bool) SetConfigRequest
+
+	GetConfig(dsts ...interface{}) error
+}
```
milanlenco: What goes into `dsts ...interface{}`?

plugins/dispatcher/dispatcher.go
```diff
@@ -0,0 +1,299 @@
...
+			p.store.Update(kv.GetKey(), kv.(*ProtoWatchResp).Val)
+		}
+	}
+
+	seqID, err := txn.Commit(ctx)
+	if err != nil {
+		p.Log.Errorf("Transaction %d failed: %v", seqID, err)
+		return err, nil
+	} else {
+		if txErr, ok := err.(*kvs.TransactionError); ok && len(txErr.GetKVErrors()) > 0 {
```
milanlenco: The processing of TransactionError should be under the `if` branch above.

ondrej-fabry: My mistake. It was first made when the Commit returned two ways for errors.

PR #1101, opened 2019-01-07
[WIP, FOR REVIEW] kvscheduler: value status

milanlenco:
Proposal for kvscheduler API in regards to value status. The representation is based on protobuf so that it can be directly serialized and saved/sent somewhere.


plugins/kvscheduler/api/value_status.proto
```diff
@@ -0,0 +1,63 @@
...
+
+message ValueStatus {
+    string key = 1;
+    ValueState state = 2;
+    ValueError error = 3;
+
+    map<string, DerivedValueStatus> derived = 4; // key -> status of a derived value
+}
+
+message DerivedValueStatus {
```
ondrej-fabry: Why introduce new type for derived values. The value and derived value share same structure, so it might be better to share the type and define it like this:

```proto
message Value {
    string key = 1;
    ValueState state = 2;
    ValueError error = 3;
}

message ValueStatus {
    Value value = 1;
    repeated Value derived_values = 2;
}
```

plugins/kvscheduler/api/value_status.proto
```diff
@@ -0,0 +1,63 @@
...
+    // be configured because there is no registered descriptor associated with it.
+    UNIMPLEMENTED = 5;
+
+    // ValueState_FAILED marks (NB) value for which the last executed operation
+    // returned an error. The error is stored in the <error> field of the value
+    // status.
+    FAILED = 6;
+}
+
+message ValueError {
```
ondrej-fabry: This should rather be named `ValueDetail`, because the value state might be in non-failed state like `PENDING` and `ValueDetail` could describe the missing dependency, or the state could be in `INVALID` state and this could describe what part of it is not valid.

plugins/kvscheduler/api/txn_options.go
```diff
@@ -34,6 +34,11 @@ const (
 	// the context.
 	retryCtxKey
 
+	// default parameters for the *retry* txn option
+	defaultRetryPeriod = time.Second
```
ondrej-fabry: These could perhaps be exported variables instead of constants to allow changing the defaults globally.

plugins/kvscheduler/api/value_status.proto
```diff
@@ -0,0 +1,78 @@
...
+    UNDEFINED = 0;
+    VALIDATE = 1;
+    ADD = 2;
+    MODIFY = 3;
+    DELETE = 4;
+    UPDATE = 5;
+}
+
+message ValueStatus {
+    message Value {
```
ondrej-fabry: The `Value` should be defined outside of `ValueStatus`. The variable name will be shorter (`ValueStatus_Value` vs. `Value`) and proto will be easier to read.

PR #1105, opened 2019-01-07
Nat session affinity support

lukasmacko:
Signed-off-by: Lukas Macko <lmacko@cisco.com>


plugins/vppv2/natplugin/vppcalls/nat_vppcalls.go
```diff
@@ -341,6 +341,7 @@ func (h *NatVppHandler) handleNat44StaticMappingLb(mapping *nat.DNat44_StaticMap
 		SelfTwiceNat: boolToUint(mapping.TwiceNat == nat.DNat44_StaticMapping_SELF),
 		Out2inOnly:   1,
 		IsAdd:        boolToUint(isAdd),
+		Affinity:     uint32(boolToUint(mapping.SessionAffinity)),
```
milanlenco: dump_nat_vppcalls.go needs to be also updated to include dumping of the affinity

rastislavszabo: I think `Affinity` is not a boolean, but a timer. 

We need to have the timer configurable, as per https://kubernetes.io/docs/concepts/services-networking/service/ :

Any traffic bound for the Service’s IP:Port is proxied to an appropriate backend without the clients knowing anything about Kubernetes or Services or Pods. Client-IP based session affinity can be selected by setting service.spec.sessionAffinity to “ClientIP” (the default is “None”), and you can set the max session sticky time by setting the field service.spec.sessionAffinityConfig.clientIP.timeoutSeconds if you have already set service.spec.sessionAffinity to “ClientIP” (the default is “10800”)

lukasmacko: ah, I'll fix it.

PR #1143, opened 2019-01-30
POC: logically grouped configuration items.

milanlenco:
This is just a POC code to spark a further discussion on assigning user-defined scheduler-opaque logical meta information to key-value pairs. Here I have just played with what I called `groupings` - a slice of potentially nested groups a value belongs to. I have used slice instead of map because it gives me an order that can be used to represent nesting.

For a complex project like Contiv, we can use groupings to define logical clusters for a more readable graph visualization. For example, configuration items grouped by pods: https://pasteboard.co/HYQMOTT.png. The grouped items in the example use groupings: `["Pods", <pod-name>]` (i.e. every pod is a separate cluster nested inside "Pods", which is nested inside the main cluster).

My suggestion is to consider extending value definition to:
```
Config proto.Message
State proto.Message (state data + what is currently known as metadata)
Meta []string (or map if we can easily define nesting)
```

### DIFF
--- a/plugins/kvscheduler/api/kv_scheduler_api.go
+++ b/plugins/kvscheduler/api/kv_scheduler_api.go
@@ -262,7 +262,7 @@ type KVScheduler interface {
 type Txn interface {
 	// SetValue changes (non-derived) value.
 	// If <value> is nil, the value will get deleted.
-	SetValue(key string, value proto.Message) Txn
+	SetValue(key string, value proto.Message, groupings []string) Txn
 
 	// Commit orders scheduler to execute enqueued operations.
 	// Operations with unmet dependencies will get postponed and possibly
--- a/plugins/kvscheduler/graphviz.go
+++ b/plugins/kvscheduler/graphviz.go
@@ -14,6 +14,7 @@ import (
 	"strings"
 	"text/template"
 	"time"
+	"math/rand"
 
 	kvs "github.com/ligato/vpp-agent/plugins/kvscheduler/api"
 	"github.com/ligato/vpp-agent/plugins/kvscheduler/internal/graph"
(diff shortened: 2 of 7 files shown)


plugins/kvscheduler/api/kv_scheduler_api.go
```diff
@@ -262,7 +262,7 @@ type KVScheduler interface {
 type Txn interface {
 	// SetValue changes (non-derived) value.
 	// If <value> is nil, the value will get deleted.
-	SetValue(key string, value proto.Message) Txn
+	SetValue(key string, value proto.Message, groupings []string) Txn
```
ondrej-fabry: To provide more future-proof API the groups should be set via generic metadata map.

For example:
```go
SetValue(key string, val proto.Message, metadata map[string]string) Txn
// or
SetValue(key string, val proto.Message, metadata map[string]interface{}) Txn
```

milanlenco: Both looks good to me.

To summarize and further generalize, I would propose that each graph node carries 4 fields, comprising what is currently known as "value":

```
IntendedConfig proto.message (currently hidden inside of the *LastUpdateFlag*)
ActualConfig proto.message
StateData proto.message (can be changed by a CRUD operation or through a notification)
MetaData map[string]interface{} / map[string][]string
```

At least one of the proto.Message must be non-nil for the node to be kept in the graph.

I would like to take out the indeded config from the LastUpdateFlag and make it an integral part of the "value". The flags should only carry run-time attributes of basic data-types, aiding the scheduling algorithm, but certainly not hiding proto messages.

`StateData` would consist of what is currently stored to `metadata` + some stuff that changes asynchronously, outside of CRUD operations, such as interface link status. The idea would be that state data can be edited by CRUD methods and via notifications. But since notification may want to alter only a subset of state data, we will probably have to delegate notification effects via callbacks, e.g.:
`PushSBNotification(key string, func (currentStateData proto.Message) newStateData proto.Message)`
Any alternative ideas are welcomed...

For most objects only configuration will be defined and `StateData` will be left as nil (e.g. routes, arps, dnats). Interfaces are the most complex items and will use both config and state data.
In few cases, such as microservices, the configuration is actually undefined and only  `StateData` will be received via `Retrieve` (renamed `Dump`) or through notifications.

PR #1152, opened 2019-02-05
Implement multi source for data in orchestrator

ondrej-fabry:
This PR implements support for multiple sources for data in orchestrator.

It currently supports combining RPC+KVDB source while support combined resync for them.


plugins/orchestrator/orchestrator.go
```diff
@@ -0,0 +1,187 @@
...
+
+	return nil
+}
+
+// InitialSync will start initial synchronization with downstream.
+func (p *Plugin) InitialSync() {
+	p.Log.Debugf("starting initial sync")
+	txn := p.KVScheduler.StartNBTransaction()
+	ctx := kvs.WithResync(context.Background(), kvs.DownstreamResync, true)
+	if _, err := txn.Commit(ctx); err != nil {
```
milanlenco: Will this not remove everything configured after agent restart?

milanlenco:
Looks good for the start. In Contiv we also persist and keep up-to-date GRPC-received config in Bolt DB. Then if agent restarts, we are able to resync with both KVDB and GRPC input data without requiring GRPC clients to watch agent status and trigger resync.

plugins/orchestrator/orchestrator.go
ondrej-fabry: Is there a way to sync with SB without removing everything configured?? I was under impression that:
- FullResync: syncs NB-KVS-SB (comparing NB-SB and fixing differences)
- DownstreamResync: syns KVS-SB (not touching anything in SB, just dumping to update)
- UpstreamResync: syncs NB-KVS? (not sure here)

PR #1174, opened 2019-02-15
Updated Segment Routing to Ligato v2 changes

fgschwan:
updated code + unit tests + robot tests
~~currently forcing usage of vpp 19.01 due to currently bad handling of vpp switching (all changes for that in separate commit)~~ VPP detection fixed in https://github.com/ligato/vpp-agent/pull/1174/commits/047e3eb5a73a66bdeb9f6c443eefbef9c1cb63cf


plugins/vpp/srplugin/descriptor/localsid.go
```diff
@@ -0,0 +1,282 @@
...
+		UpdateWithRecreate: d.UpdateWithRecreate,
+		Dependencies:       d.Dependencies,
+	}
+}
+
+// EquivalentLocalSIDs determines whether 2 localSIDs are logically equal. This comparison takes into consideration also
+// semantics that couldn't be modeled into proto models (i.e. SID is IPv6 address and not only string)
+func (d *LocalSIDDescriptor) EquivalentLocalSIDs(key string, oldLocalSID, newLocalSID *srv6.LocalSID) bool {
+	if oldLocalSID == nil || newLocalSID == nil {
+		return oldLocalSID == newLocalSID
```
milanlenco: nil checks are not needed - if the previous or the new value is nil, the operation to execute will be Create or Delete.  ValueComparator is only used for Update, therefore it will not be called in those cases.

plugins/vpp/srplugin/descriptor/localsid.go
```diff
@@ -0,0 +1,282 @@
...
+	sid, _ := ParseIPv6(value.GetSid()) // already validated
+	if err := d.srHandler.DeleteLocalSid(sid); err != nil {
+		return errors.Errorf("failed to delete local sid %s: %v", sid.String(), err)
+	}
+	return nil
+}
+
+// UpdateWithRecreate always returns true - LocalSIDs are always updated via re-creation.
+func (d *LocalSIDDescriptor) UpdateWithRecreate(key string, oldValue, newValue *srv6.LocalSID, metadata interface{}) bool {
+	return true
```
milanlenco: With the latest vpp-agent, if the Update operation is not implemented, it is automatically determined to apply any change via re-creation - i.e. it is not needed to define UpdateWithRecreate always returning true.

plugins/vpp/srplugin/vppcalls/vpp1901/srv6.go
```diff
@@ -0,0 +1,598 @@
...
+		return err
+	}
+	if len(strings.TrimSpace(string(data))) > 0 {
+		return fmt.Errorf("addition of dynamic segment routing proxy failed by returning nonblank space text in CLI: %v", string(data))
+	}
+	return nil
+}
+
+// interfaceNameMapping dumps from VPP internal names of interfaces and uses them to produce mapping from ligato interface names to vpp internal names.
+func (h *SRv6VppHandler) interfaceNameMapping() (map[string]string, error) {
```
milanlenco: The mapping between interface names and their IDs is already exposed by the ifplugin: https://github.com/ligato/vpp-agent/blob/master/plugins/vpp/ifplugin/ifplugin_api.go#L27
Example where it is being used: https://github.com/ligato/vpp-agent/blob/master/plugins/vpp/l3plugin/vppcalls/arp_vppcalls.go#L27-L30

api/models/vpp/srv6/srv6.proto
```diff
@@ -0,0 +1,123 @@
...
+message Policy {
+    string bsid = 1;                     /* binding SID (IPv6 Address) */
+    uint32 fib_table_id = 2;             /* ID of FIB table where to install Policy entry */
+    bool srh_encapsulation = 3;          /* are SR headers handled by encapsulation? (no means insertion of SR headers) */
+    bool spray_behaviour = 4;            /* spray(multicast) to all policy segments? (no means to use PolicySegment.weight to loadbalance traffic)*/
+}
+
+/* Model for segment list of SRv6 policy */
+message PolicySegmentList {
+    string policy_bsid = 1;             /* reference to policy where this policy segment belongs */
```
milanlenco: Model PolicySegmentList doesn't mention any name, yet the key is composed of it -- apparently for this reason ModelPolicySegmentList couldn't be used to generate and parse keys (hence the method ParsePolicySegmentList, the sort of which we want to get rid of). Would it be possible to add the logical name here and register model with template combining the name with the policy bsid?

fgschwan: The logical name is there only because key must be unique (name + policy_bsid = unique). The proper unique combination (from proto model fields only) would be from all fields (policy_bsid + wight + segments), but that could be possible really long key (segments field is list and it's length is not limited by segment routing standards). Therefore i prefer the logical name, but i don't want to pollute the model with it. I didn't know that you wanted to get rid of all custom parse methods, but now it seems to me like a good trade-off (1 custom parse method vs polluted model).

plugins/vpp/srplugin/descriptor/localsid.go
fgschwan: fixed in https://github.com/ligato/vpp-agent/pull/1174/commits/63aa9e15d7b692c62577c288db87eac9c802974f

plugins/vpp/srplugin/descriptor/localsid.go
fgschwan: fixed in https://github.com/ligato/vpp-agent/pull/1174/commits/18b2bc5124634cb1720353d988f47e0caea00f18

milanlenco:
IMHO the implementation is quite complicated and doesn't really fit very well into the kvscheduler framework. The use of the internal cache brings bad memories of v1 plugins. The (premature) optimization to split the lists from policies not only introduced cyclic dependency, but also made the models quite fragmented: policy is completely bare - to learn about given policy from NB config, one has to go through all the segment lists and filter out those that reference other policies, and then continue through sid-s into the corresponding LocalSID definitions (if any) to get a full picture of the policy. It is true, however, that in some cases Update operation may be more efficient between client and the agent NB (between agent and VPP it doesn't matter), but in others in may result in many Put/Delete operations: for example, in Contiv one policy will implement one service, where for every backend there will be one segment list. With 100 backends, (un)deploying the service will cost 100+1 put/delete operations (and consequent notifications from KVDB if used). The latency introduced by the higher number of operations will cost much more than slightly bigger message payload. Also, having one key-value pair for each segment list will make the life of clients much more difficult - they will have to try harder to ensure that everything is in-sync in NB, that there are no dangling lists, etc.

I don't want to enforce my opinion (and will not delay the merge), however, it is just that in agent v1 we already tried optimizations of this kind - e.g. splitting ACLs into separate rules, which only resulted in complicated and buggy implementation without measurable improvements, so we decided to get rid of it. DNATs are very similar to SR6 policies in that they also match IPs with load-balanced next-hop destinations (used to implement services in Contiv for IPv4), and we didn't notice any performance issues in listing all the backends inside one model (per service)...

api/models/vpp/srv6/srv6.proto
milanlenco: It has to be possible to generate key using only the content of the value - i.e. the Name must be included - otherwise GRPC/local client will not work correctly. It is because the values to configure are given to the client without keys (unlike when KVDB is used as NB interface) and the client uses the models to build the appropriate key for every value, e.g.: https://github.com/ligato/vpp-agent/blob/master/client/local_client.go#L53

fgschwan: Ok, i have to change it so that key can be generated from value. For PolicySegmentList it means to add name to model (merge of policy and policysegmentlist as suggested https://github.com/ligato/vpp-agent/pull/1174#issuecomment-464697745 would also fix this). I also have the same problem in steering (https://github.com/ligato/vpp-agent/pull/1174/files#diff-60062a829b92c1b0f46b0d677ae8764bR103) where i use in key name that is not in value. I will add the name into steering model too.

fgschwan: Name moved into models in https://github.com/ligato/vpp-agent/pull/1174/commits/edd20eea90f66870a18d336731c8de81b2365019 . 
However this doesn't removes ParsePolicySegmentList mentioned in first comment in this thread, because there is no standard way how to extract policyBSID from PolicySegmentList key (my first usage of that method) and also there is no proper name validation (my second usage) (https://github.com/ligato/vpp-agent/blob/42cf54a33060b64cd4b8ff3c7418a4e1798bc9d6/pkg/models/spec.go#L87)

fgschwan:
> IMHO the implementation is quite complicated and doesn't really fit very well into the kvscheduler framework. The use of the internal cache brings bad memories of v1 plugins. The (premature) optimization to split the lists from policies not only introduced cyclic dependency, but also made the models quite fragmented: policy is completely bare - to learn about given policy from NB config, one has to go through all the segment lists and filter out those that reference other policies, and then continue through sid-s into the corresponding LocalSID definitions (if any) to get a full picture of the policy. It is true, however, that in some cases Update operation may be more efficient between client and the agent NB (between agent and VPP it doesn't matter), but in others in may result in many Put/Delete operations: for example, in Contiv one policy will implement one service, where for every backend there will be one segment list. With 100 backends, (un)deploying the service will cost 100+1 put/delete operations (and consequent notifications from KVDB if used). The latency introduced by the higher number of operations will cost much more than slightly bigger message payload. Also, having one key-value pair for each segment list will make the life of clients much more difficult - they will have to try harder to ensure that everything is in-sync in NB, that there are no dangling lists, etc.
> 
> I don't want to enforce my opinion (and will not delay the merge), however, it is just that in agent v1 we already tried optimizations of this kind - e.g. splitting ACLs into separate rules, which only resulted in complicated and buggy implementation without measurable improvements, so we decided to get rid of it. DNATs are very similar to SR6 policies in that they also match IPs with load-balanced next-hop destinations (used to implement services in Contiv for IPv4), and we didn't notice any performance issues in listing all the backends inside one model (per service)...

Ok, i will merge it to one big json (1 policy descriptor handling policy and policy segment lists)

fgschwan:
Policy and Segment Lists are merged now, see https://github.com/ligato/vpp-agent/pull/1174/commits/05e4a8025104b093da1a85feb4f6123146bea857

PR #1181, opened 2019-02-21
Comment installing debugger - ARM64 version of go debugger does not exists

stanislav-chlebec:


docker/dev/Dockerfile
```diff
@@ -89,9 +89,11 @@ RUN mkdir -p "$GOPATH/src" "$GOPATH/bin" \
 
 # install debugger
 RUN \
+      if [ "$GOLANG_OS_ARCH" == "linux-amd64" ]; \
+      then \
 	go get -u github.com/go-delve/delve/cmd/dlv && dlv version; \
```
ondrej-fabry: This is the command which installs the debugger which I believe will work for ARM.

docker/dev/Dockerfile
```diff
@@ -89,9 +89,11 @@ RUN mkdir -p "$GOPATH/src" "$GOPATH/bin" \
...
-	curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh && dep version;
-
+	curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh && dep version; \
```
ondrej-fabry: This is a call that installs dependency manager `dep`. Which actually could be replaced with `go get -u github.com/golang/dep/cmd/dep` and could work even for ARM.

PR #1200, opened 2019-02-28
Handle unused (non-default) plugin (without panic).

milanlenco:
This is a quick solution to avoid panic when some VPP plugin we support is not loaded. It requires that our counterpart plugin is not loaded as well. This PR makes sure that both configurator and restapi plugin can handle a missing (non-default) plugin.

A dynamic and more robust solution will most likely require to introduce new value state - `UNSUPPORTED`.

### DIFF
--- a/plugins/configurator/dump.go
+++ b/plugins/configurator/dump.go
@@ -1,6 +1,7 @@
 package configurator
 
 import (
+	"github.com/go-errors/errors"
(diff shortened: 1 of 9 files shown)


milanlenco:
Have to merge to unblock Vinay, leaving potential discussion for tomorrow...

PR #1201, opened 2019-03-01
kvscheduler docs: rephrase explanation for derived values, add diagrams

milanlenco:


ondrej-fabry:
It seems to me that the large arrow in [the image for derived values](https://github.com/ligato/vpp-agent/blob/590c024a3cb0d27aba2a6befcc646b2d09a2fdbc/docs/kvscheduler/img/derived-values.svg) is missing "values" at the end?

ondrej-fabry:
In the mentioned picture, shouldn't the left side be named "VPP Bridge Domain Model" instead of "VPP Bridge Domain descriptor" and also not show the dependencies? Because it actually describes how we break it down, but we break it down from model and not another descriptor.. and this might confuse someone.

milanlenco:
I guess you are right - originally I wanted to show differences between using and not using derived values for bridge domains, but the end result was really just confusing. I have simplified the diagram - now it only shows the actual solution.

PR #1207, opened 2019-03-05
Linux loopback support

lukasmacko:
Signed-off-by: Lukas Macko <lmacko@cisco.com>


api/models/linux/interfaces/interface.proto
```diff
@@ -48,3 +50,7 @@ message VethLink {
 message TapLink {
     string vpp_tap_if_name = 1;     /* Logical name of the VPP TAP interface (mandatory for TAP_TO_VPP) */
 };
+
+message LoopLink {
```
VladoLavor: Not needed since there are no data fields

plugins/linux/ifplugin/descriptor/interface.go
```diff
@@ -260,6 +270,22 @@ func (d *InterfaceDescriptor) Validate(key string, linuxIf *interfaces.Interface
...
+		if err != nil {
+			return kvs.NewInvalidValueError(ErrNamespaceDoesntExists, "namespace")
+		}
+		nsHandle.Close()
+
+		// in each network namespace there is exactly one loopback interface,
+		// multiple configurations for a single namespaces is not valid
+		alias, err := d.getLoopbackAlias(nsCtx, linuxIf)
+		if err != nil || (alias != "" && alias != linuxIf.Name) {
+			d.log.WithField("error", err).Errorf("loopback already configured using logical name '%v'", alias)
```
milanlenco: This error is I think also returned when the loopback doesn't exist (if it can even happen), would it be possible to differentiate these two cases?

plugins/linux/ifplugin/descriptor/interface.go
```diff
@@ -260,6 +270,22 @@ func (d *InterfaceDescriptor) Validate(key string, linuxIf *interfaces.Interface
 		if linuxIf.GetVeth().GetPeerIfName() == "" {
 			return kvs.NewInvalidValueError(ErrVETHWithoutPeer, "peer_if_name")
 		}
+	case *interfaces.Interface_Loop:
+		nsCtx := nslinuxcalls.NewNamespaceMgmtCtx()
+		nsHandle, err := d.nsPlugin.GetNamespaceHandle(nsCtx, linuxIf.Namespace)
+		if err != nil {
+			return kvs.NewInvalidValueError(ErrNamespaceDoesntExists, "namespace")
```
VladoLavor: This can be misleading. `GetNamespaceHandle()` also tries to create the namespace if missing (calls `getOrCreateNs()` inside) and returned error can mean that the namespace failed to be created or read, or something else.

Perhaps the namespace check can be moved to the `getLoopbackAlias` and use some error like 'LoopbackValidationError' for all cases (failed get/create namespace, failed switch to namespace, alias already exists) and log the actual case.

api/models/linux/interfaces/interface.proto
```diff
@@ -48,3 +50,7 @@ message VethLink {
...
+
```
milanlenco: In vpp/ifplugin, we didn't define empty "link" for interfaces that do not have any type-specific attributes: https://github.com/ligato/vpp-agent/blob/master/api/models/vpp/interfaces/interface.proto#L60-L68 (for example SOFTWARE_LOOPBACK ). I think it is less confusing for users, but it is up to the discussion which is better.

ondrej-fabry: If there are no plans to add anything it's probably better to avoid it. Less unnecessary work.

plugins/linux/ifplugin/descriptor/interface_loop.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	return metadata, nil
+}
+
+// deleteLoopback clear associated interface alias
+func (d *InterfaceDescriptor) deleteLoopback(nsCtx nslinuxcalls.NamespaceMgmtCtx, linuxIf *interfaces.Interface) error {
+	hostName := getHostIfName(linuxIf)
+
+	revert, err := d.nsPlugin.SwitchToNamespace(nsCtx, linuxIf.Namespace)
```
VladoLavor: This step can be omitted if the loopback is in default namespace if it is a valid case (also for delete)

plugins/linux/ifplugin/descriptor/interface.go
```diff
@@ -261,6 +271,24 @@ func (d *InterfaceDescriptor) Validate(key string, linuxIf *interfaces.Interface
 			return kvs.NewInvalidValueError(ErrVETHWithoutPeer, "peer_if_name")
 		}
 	}
+
+	if linuxIf.Type == interfaces.Interface_LOOPBACK {
+		nsCtx := nslinuxcalls.NewNamespaceMgmtCtx()
```
milanlenco: I just realized that inside the scheduling algorithm, configuration is validated before the dependencies are checked (it was really just meant for the validation of the configuration, not the environment). For loopback configuration with microservice namespace, it is a valid case if the namespace doesn't exist yet (it will be pending and created when the notifications about microservice arrives). Therefore it may be better to put these checks into Create (and mark the "already-exists" error as unsuitable for Retry: https://github.com/ligato/vpp-agent/blob/master/plugins/kvscheduler/api/kv_descriptor_api.go#L212-L217). But for Contiv this is not relevant (and loopbacks are not used beyond that), so I guess we can also leave it as is and just to keep that scenario in mind.

ondrej-fabry:
Can this be merged?

lukasmacko:
I think it can, I was just waiting for @milanlenco to approve the implemented changes. However, current solution is working, no open issuesm further improvements can be done in a subsequent PR.

milanlenco:
LGTM

PR #1212, opened 2019-03-06
Fix and re-enable remaining kvscheduler UTs.

milanlenco:


plugins/kvscheduler/txn_exec.go
```diff
@@ -200,6 +200,15 @@ func (s *Scheduler) applyValue(args *applyValueArgs) (executed kvs.RecordedTxnOp
 		// inherit retry arguments from the last NB txn for this value
 		lastUpdateFlag.retryEnabled = prevUpdate.retryEnabled
 		lastUpdateFlag.retryArgs = prevUpdate.retryArgs
+	} else if args.isDerived {
```
ondrej-fabry: What does this do?

milanlenco: This is for some rather special scenarios, in which value entered via NB transaction remains pending, but becomes ready and gets Created with a SB notification. But if the Create operation fails, we need to determine whether to schedule Retry - this should be determined from the txn options of the original NB txn. Therefore we need to remember txn options alongside the value using last-update flag. Derived values then inherit these options from their "parent" values.

PR #1214, opened 2019-03-07
kvgraph performance improvements

milanlenco:
Profiling clearly shows that there are some performance issues in the  kvscheduler/graph, and they become more and more significant as the size of the graph increases.
I will investigate and try to come up with some improvements.

NOTE: too much of copying is probably not the only issue...


milanlenco:
It is much better now, however there are still things that can be improved. A major bottleneck now are the `AnyOf` dependency selectors - for a newly inserted node, the kvgraph has to call every `AnyOf` selector of every existing node to test if there is an edge going into the new node. The callback is too much of a black box for the graph, so perhaps we could extend `AnyOf` and allow to combine callback with key prefix(es) that the target key must have (multiple key prefixes would be ORed and then ANDed with the selector if defined), in order to decrease the set of candidates. In all of our use cases of the AnyOf callback across linux and vpp plugins, the set of candidates can be quite significantly narrowed down using key prefix(es), only the suffix requires parsing and processing with a callback.

PR #1232, opened 2019-03-14
Linux iptables plugin implementation

rastislavszabo:


plugins/linux/iptablesplugin/descriptor/rulechain.go
```diff
@@ -0,0 +1,418 @@
...
+}
+
+// Dependencies lists dependencies for a iptables rule chain.
+func (d *RuleChainDescriptor) Dependencies(key string, rch *linux_iptables.RuleChain) []kvs.Dependency {
+	if len(rch.Interfaces) > 0 {
+		// the associated interfaces must exist
+		var deps []kvs.Dependency
+		for _, i := range rch.Interfaces {
+			deps = append(deps, kvs.Dependency{
+				Label: ruleChainInterfaceDep,
```
milanlenco: label must be unique for every dependency - add interface name as suffix, e.g.: https://github.com/ligato/vpp-agent/blob/master/plugins/vpp/natplugin/descriptor/nat44_dnat.go#L236

plugins/linux/iptablesplugin/descriptor/rulechain.go
```diff
@@ -0,0 +1,418 @@
...
+type retrievedRuleChains struct {
+	chains []adapter.RuleChainKVWithMetadata
+	err    error
+}
+
+// Retrieve returns all iptables rule chain entries managed by this agent.
+func (d *RuleChainDescriptor) Retrieve(correlate []adapter.RuleChainKVWithMetadata) ([]adapter.RuleChainKVWithMetadata, error) {
+	var values []adapter.RuleChainKVWithMetadata
+
+	goRoutinesCnt := len(correlate) / minWorkForGoRoutine
```
milanlenco: Retrieve every namespace only once or there will be duplicities if multiple chains for the same namespace are in correlate, e.g.: https://github.com/ligato/vpp-agent/blob/master/plugins/linux/ifplugin/descriptor/interface.go#L613-L621

plugins/linux/iptablesplugin/descriptor/rulechain.go
```diff
@@ -0,0 +1,418 @@
...
+
+	// parallelization of the Retrieve operation
+	goRoutinesCnt int
+}
+
+// NewRuleChainDescriptor creates a new instance of the iptables RuleChain descriptor.
+func NewRuleChainDescriptor(
+	scheduler kvs.KVScheduler, ipTablesHandler linuxcalls.IPTablesAPI, nsPlugin nsplugin.API,
+	log logging.PluginLogger, goRoutinesCnt int) *RuleChainDescriptor {
+
```
milanlenco: Recently we have come up with a cleaner layout for descriptor constructor: https://github.com/ligato/vpp-agent/blob/master/examples/kvscheduler/mock_plugins/ifplugin/descriptor/interface.go#L64-L92
It avoids the GetDescriptor method and registration inside the plugin is cleaner (plus it doesn't have to import adapter, which should be really only internal to descriptor): https://github.com/ligato/vpp-agent/blob/master/examples/kvscheduler/mock_plugins/ifplugin/ifplugin.go#L76-L81

plugins/linux/iptablesplugin/descriptor/rulechain.go
```diff
@@ -0,0 +1,418 @@
...
+		Dependencies:         d.Dependencies,
+		RetrieveDependencies: []string{ifdescriptor.InterfaceDescriptorName},
+	}
+}
+
+// EquivalentRuleChains is a comparison function for two RuleChain entries.
+func (d *RuleChainDescriptor) EquivalentRuleChains(key string, oldRCh, newRch *linux_iptables.RuleChain) bool {
+
+	// first, compare everything except the rules
+	oldCopy := proto.Clone(oldRCh).(*linux_iptables.RuleChain)
```
milanlenco: I'm afraid that copying may have an impact on performance since the comparison function is called quite often.

plugins/linux/iptablesplugin/descriptor/rulechain.go
```diff
@@ -0,0 +1,418 @@
...
+		}
+	}
+
+	return nil
+}
+
+// UpdateWithRecreate tells the scheduler that the update should be performed with re-create.
+func (d *RuleChainDescriptor) UpdateWithRecreate(key string, oldRCh, newRCh *linux_iptables.RuleChain, oldMetadata interface{}) bool {
+	// always update with recreate
+	return true
```
milanlenco: UpdateWithRecreate is not needed here - without Update, recreation is the default (and only possible) behaviour

PR #1236, opened 2019-03-15
kvgraph: optimized lookup over graph edges

milanlenco:
~~TODO: add key prefixes alongside AnyOf selectors to all descriptors where it is used~~

I realized that the solution is not very efficient in one place and will need to rethink it when the time permits, leaving WIP for the time being...


ondrej-fabry:
@milanlenco thanks Milan, This PR seems to greatly improve the performance! For perf test with 10k tunnels it takes only **~22.5%** (10m -> 2m15s) of time than before.

PR #1246, opened 2019-03-26
Fix retrieving VPP metrics in telemetry plugin

ondrej-fabry:

### DIFF
--- a/plugins/telemetry/prometheus.go
+++ b/plugins/telemetry/prometheus.go
@@ -209,14 +209,11 @@ func (p *Plugin) registerPrometheus() error {
 	return nil
 }
 func (p *Plugin) updatePrometheus() {
-	p.Log.Debugf("updating")
-
 	// Update runtime
 	runtimeInfo, err := p.handler.GetRuntimeInfo()
 	if err != nil {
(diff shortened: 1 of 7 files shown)


plugins/telemetry/prometheus.go
```diff
@@ -205,11 +209,14 @@ func (p *Plugin) registerPrometheus() error {
 	return nil
 }
 func (p *Plugin) updatePrometheus() {
+	p.Log.Debugf("updating")
```
VladoLavor: This does not tell much.

plugins/telemetry/prometheus.go
```diff
@@ -205,11 +209,14 @@ func (p *Plugin) registerPrometheus() error {
...
+
 	// Update runtime
 	runtimeInfo, err := p.handler.GetRuntimeInfo()
 	if err != nil {
-		p.Log.Errorf("Command failed: %v", err)
+		p.Log.Errorf("GetRuntimeInfo failed: %v", err)
 	} else {
+		//p.Log.Debugf("runtime info: %+v", runtimeInfo)
```
VladoLavor: I guess this can be removed (same for other commented lines)

PR #1249, opened 2019-03-26
Added support for SRv6 end function End.DT4 and End.DT6 + other SRv6 fixes

fgschwan:
Main goal: 
1. support for End.DT4, End.DT6 end functions + unit tests + robot test (crud only)

Secondary goals: 
1. fix failures in current robot tests for SRv6
2. extend robot tests for Srv6 to cover all implemented end functions (current test covers only CRUD for LocalSID with End.DX6 end function)
3. fix problems surfaced while doing robot test from 2. secondary goal (some end functions were never used -> fixed their wrong modeling in NB API and implementation)


api/models/vpp/l3/keys.go
```diff
@@ -100,3 +100,8 @@ func ParseProxyARPInterfaceKey(key string) (iface string, isProxyARPInterfaceKey
 	}
 	return "", false
 }
+
+// RoutePrefix returns longest-common prefix of keys representing route that is written to given vrf table.
+func RoutePrefix(vrf string) string {
```
ondrej-fabry: The `vrf` parameter should be `uint32` just like the `VrfId` filed for `Route`: https://github.com/ligato/vpp-agent/blob/f94992578ccc2289b5c123f83e98600699016c52/api/models/vpp/l3/route.pb.go#L50
This will eliminate the need to convert it from every caller.

fgschwan: fixed in https://github.com/ligato/vpp-agent/pull/1249/commits/c56066ed17628a796a1a065c609c30a16f01c1e8

PR #1260, opened 2019-04-03
Run supervisord as process with PID 1

matrohon:
Signed-off-by: Mathieu Rohon <mathieu.rohon@orange.com>

### DIFF
--- a/docker/dev/Dockerfile
+++ b/docker/dev/Dockerfile
@@ -123,4 +123,4 @@ WORKDIR /
 
 # run supervisor as the default executable
 CMD rm -f /dev/shm/db /dev/shm/global_vm /dev/shm/vpe-api && \
-    /usr/bin/supervisord -c /etc/supervisord/supervisord.conf
+    exec /usr/bin/supervisord -c /etc/supervisord/supervisord.conf
--- a/docker/prod/Dockerfile
+++ b/docker/prod/Dockerfile
@@ -65,4 +65,4 @@ WORKDIR /root/
 
 # run supervisor as the default executable
 CMD rm -f /dev/shm/db /dev/shm/global_vm /dev/shm/vpe-api && \
-    /usr/bin/supervisord -c /etc/supervisord/supervisord.conf
+    exec /usr/bin/supervisord -c /etc/supervisord/supervisord.conf


matrohon:
Running supervisord as process with Pid 1 allows it to receive signals from kubelet. Then, it can send SIGTERM to processes it manages, for them to terminate gracefully.

It prevents issues such as https://github.com/networkservicemesh/networkservicemesh/issues/894

docker/dev/Dockerfile
```diff
@@ -121,6 +121,7 @@ RUN cd $GOPATH/src/github.com/ligato/vpp-agent \
 
 WORKDIR /
 
+RUN rm -f /dev/shm/db /dev/shm/global_vm /dev/shm/vpe-api
```
ondrej-fabry: Executing this command as `RUN` instruction will render its entire purpose useless, because `RUN` instructions are only executed when building an image and these files do not exist at that point yet.

The purpose of this command is to forcefully remove files that might remain there after potential VPP crash, in such situation these files are not cleaned up by VPP, which are used for communication via shared memory (binary API) and we need to remove them before starting VPP again, otherwise connection will fail.

By putting this command in the `RUN` instruction, the files will not be removed if stopped container is started again after VPP crash.

docker/dev/Dockerfile
```diff
@@ -121,6 +121,7 @@ RUN cd $GOPATH/src/github.com/ligato/vpp-agent \
...
+
 # run supervisor as the default executable
-CMD rm -f /dev/shm/db /dev/shm/global_vm /dev/shm/vpe-api && \
-    /usr/bin/supervisord -c /etc/supervisord/supervisord.conf
+ENTRYPOINT ["/usr/bin/supervisord","-c","/etc/supervisord/supervisord.conf"]
```
ondrej-fabry: By using `ENTRYPOINT` instruction for starting supervisord you are forcing developers that want to run the container with shell only (no supervisord at all, perhaps for some debugging purposes) to override the entrypoint with `--entrypoint` flag instead of just appending custom command in the `docker run`.

`docker run -it ligato/vpp-agent /bin/bash` vs. `docker run -it --entrypoint /bin/bash ligato/vpp-agent`

docker/dev/Dockerfile
matrohon: Thanks, I better understand th purpose of those command now

edwarnicke: @ondrej-fabry How exactly are those /dev/shm/... files getting into the container to require cleanup?  If vpp crashes, the whole container goes down, and gets restarted from a fresh clean image.  /dev/shm lives inside the container image, so it goes away with the container.

docker/dev/Dockerfile
edwarnicke: +1

matrohon:
@ondrej-fabry I didn't though about using exec to replace the parent process. I just tested it, it works like a charm, thanks.

edwarnicke:
@ondrej-fabry I'm still curious how its possible for the container to have remnants of a past vpp crash in /dev/shm given that supervisord will exit the container on a vpp crash, and a restart of the container will start clean...

ondrej-fabry:
@edwarnicke In the common case in which we usually run the container it does not make much sense probably, because as you mentioned, on any crash of VPP or Agent we replace the container with new one.

However to cover other uses, for example when someone uses an option to keep supervisor running ([`RETAIN_SUPERVISOR`](https://github.com/ligato/vpp-agent/tree/master/docker/dev#starting-the-image)) and then stops the container gracefully and starts it again, the `rm` command will get rid of any files that might have remained.

This is only situation that comes to my mind.. I suspect it was added to prevent some similar case. Found the PR that introduced it: https://github.com/ligato/vpp-agent/pull/148 sadly.. not much explanation there.

PR #1262, opened 2019-04-04
Agentctl2

kozemcak:


kozemcak:
Create new agentctl2

cmd/agentctl2/cmd/show_version_cmd.go
```diff
@@ -0,0 +1,26 @@
...
+`,
+	Run: versionFunc,
+}
+
+func init() {
+	RootCmd.AddCommand(showVersion)
+}
+
+func versionFunc(cmd *cobra.Command, args []string) {
+	fmt.Printf("agentctl version 0.1\n")
```
ondrej-fabry: The version of agent can be retrieved from `ligato/cn-infra/agent` package. 

See how the vpp-agent prints it at boot: https://github.com/ligato/vpp-agent/blob/731e9b4b9b2e37407e805221dcab9ccbd0482353/cmd/vpp-agent/main.go#L41

cmd/agentctl2/utils/db_utils.go
```diff
@@ -0,0 +1,654 @@
...
+	nat "github.com/ligato/vpp-agent/api/models/vpp/nat"
+
+	linterface "github.com/ligato/vpp-agent/api/models/linux/interfaces"
+	ll3 "github.com/ligato/vpp-agent/api/models/linux/l3"
+
+	"errors"
+)
+
+const (
+	ACLPath          = "config/vpp/acls/v2/acl/"
```
ondrej-fabry: These shouldn't be hard-coded.

See example how the aclplugin retrieves key prefix: https://github.com/ligato/vpp-agent/blob/731e9b4b9b2e37407e805221dcab9ccbd0482353/plugins/vpp/aclplugin/descriptor/acl.go#L66

cmd/agentctl2/utils/db_utils.go
```diff
@@ -0,0 +1,654 @@
...
+
+// lRouteWithMD contains a data record for interface and its
+// etcd metadata.
+type lRouteWithMD struct {
+	Config *lRouteConfigWithMD
+}
+
+// VppData defines a structure to hold all etcd data records (of all
+// types) for one VPP.
+type VppData struct {
```
ondrej-fabry: There actually is type `Config` which provides all of this configuration: https://github.com/ligato/vpp-agent/blob/731e9b4b9b2e37407e805221dcab9ccbd0482353/api/configurator/configurator.proto#L11-L14

ondrej-fabry:
Please rebase on top of latest dev branch and force-push to get rid of the double commits.

cmd/agentctl2/cmd/cli_cmd.go
```diff
@@ -0,0 +1,42 @@
...
+	"fmt"
+	"os"
+
+	"github.com/ligato/vpp-agent/cmd/agentctl2/restapi"
+	"github.com/spf13/cobra"
+)
+
+// RootCmd represents the base command when called without any subcommands.
+var cliConfig = &cobra.Command{
+	Use:     "cli",
```
ondrej-fabry: Using `vppcli` might make more sense here. 

Maybe aslo, without an alias, so it's always obvious what it does, even in some scripts.

```sh
agentctl c show interfaces
# vs.
agentctl vppctl show interfaces
```

ondrej-fabry:
Let's get rid of the `_cmd` suffix for files under `cmd/agentctl/cmd/xxx_cmd.go`, because it's ridiculously redundant.

cmd/agentctl2/utils/common_utils.go
```diff
@@ -0,0 +1,217 @@
...
+
+// ParseKey parses the etcd Key for the microservice label and the
+// data type encoded in the Key. The function returns the microservice
+// label, the data type and the list of parameters, that contains path
+// segments that follow the data path segment in the Key URL. The
+// parameter list is empty if data path is the Last segment in the
+// Key.
+//
+
+func ParseKey(key string) (label string, dataType string, name string) {
```
ondrej-fabry: This shouldn't be needed.. `models` package should be able to provide this with https://github.com/ligato/vpp-agent/blob/40bd6b1899e0abe9a30da72c5430225a44cf39b2/pkg/models/models.go#L100

which returns `registeredModel` that has these methods available (even though it's unexported..): https://github.com/ligato/vpp-agent/blob/40bd6b1899e0abe9a30da72c5430225a44cf39b2/pkg/models/spec.go#L68-L108

Or you can use the specific models directly: https://github.com/ligato/vpp-agent/blob/dev/api/models/vpp/interfaces/keys.go#L30

ondrej-fabry: Also these fields are available on the `registeredModel` too.. 

(note to self: that really has to be changed.. it's weird being used unexported...)

cmd/agentctl2/utils/common_utils.go
```diff
@@ -0,0 +1,217 @@
...
+	"github.com/ligato/cn-infra/db/keyval"
+	"github.com/ligato/cn-infra/db/keyval/etcd"
+	"github.com/ligato/cn-infra/db/keyval/kvproto"
+	"github.com/ligato/cn-infra/logging"
+	"github.com/ligato/cn-infra/logging/logrus"
+	"github.com/ligato/cn-infra/servicelabel"
+)
+
+const (
+	ACLPath          = "config/vpp/acls/v2/acl/"
```
ondrej-fabry: This shouldn't be redefined. Avaiable like this: https://github.com/ligato/vpp-agent/blob/40bd6b1899e0abe9a30da72c5430225a44cf39b2/plugins/vpp/ifplugin/descriptor/interface.go#L173

cmd/agentctl2/utils/print_log.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	return Template
+}
+
+func (ll LogList) textRenderer(templates []*template.Template) (*bytes.Buffer, error) {
+	buffer := new(bytes.Buffer)
+	buffer.WriteTo(os.Stdout)
+	for _, value := range ll {
+		var wasError error
```
ondrej-fabry: I'm not sure why is this define here. Isn't the function returning if `wasError` gets some error on line 61? It could be scoped just there like this:

```go
if err := tmpl.Execute(buf, val); err != nil {
	return nil, err
}
```

ondrej-fabry: Or perhaps the `if` condition was supposed to outside of the inner loop?

cmd/agentctl2/utils/print_log.go
```diff
@@ -0,0 +1,69 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	return Template
+}
+
+func (ll LogList) textRenderer(templates []*template.Template) (*bytes.Buffer, error) {
+	buffer := new(bytes.Buffer)
+	buffer.WriteTo(os.Stdout)
```
ondrej-fabry: This is most likely doing nothing.. since the `buffer` is empty at that point.

cmd/agentctl2/utils/print_log.go
```diff
@@ -0,0 +1,69 @@
...
+
+	templates := []*template.Template{}
+
+	templates = append(templates, logLevel)
+
+	return ll.textRenderer(templates)
+}
+
+func createLogTypeTemplate() *template.Template {
+	Template, err := template.New("log").Parse(
```
ondrej-fabry: If the template is invalid for some reason it's better to sooner. You can use `Must` to achieve this: https://godoc.org/text/template#Must

Also gets parsed only once since it will be in global scope.

cmd/agentctl2/utils/print_log.go
```diff
@@ -0,0 +1,69 @@
...
+	Level  string `json:"level,omitempty"`
+}
+
+type LogList []logType
+
+func ConvertToLogList(log string) LogList {
+	data := make(LogList, 0)
+	err := json.Unmarshal([]byte(log), &data)
+
+	if nil != err {
```
ondrej-fabry: Let's start using consistent style here, preferably the common way `err != nil`. Reads weird when reversed.

cmd/agentctl2/utils/common_utils.go
kozemcak: Ok,
But I will look at this when I return from holiday.
I want to make a big refactor of this part when I return.

cmd/agentctl2/utils/print_log.go
kozemcak: This was more copy - past code. But I agree, I will change it.

PR #1273, opened 2019-04-08
Model + descriptor for VRF table

milanlenco:
TODO:
- [x] code cleanup
- [x] testing with all supported VPP versions
- [x] testing with Contiv

PR #1284, opened 2019-04-12
Temporal fix for Segment routing dependency on VRF tables

fgschwan:
This is only temporal fix that checks existence of needed vrf tables for SRv6 components. It should be later replaced by VRF model dependency (VRF model is not done yet)


milanlenco:
It seems that I will have some extra free time tomorrow to finish the pull request https://github.com/ligato/vpp-agent/pull/1273.
This temporary workaround would cause some merge conflicts with my changes and I would have to revert it then anyway. Therefore, would it be OK for you if we would not merge the workaround into the vpp-agent, and instead you would temporarily apply it directly into the vendor of Contiv?

fgschwan:
> 
> 
> It seems that I will have some extra free time tomorrow to finish the pull request #1273.
> This temporary workaround would cause some merge conflicts with my changes and I would have to revert it then anyway. Therefore, would it be OK for you if we would not merge the workaround into the vpp-agent, and instead you would temporarily apply it directly into the vendor of Contiv?

Ok, i will apply it to vendor of Contiv for now, but that means that "dep ensure" will be problematic in Contiv (it erase changes in vendor, but that can be managed in short term). I wanted to remove dependency on your pull request because i was not sure how long it would you to finish your pull request. However if you want to finish it soon then it is ok for me (ideally before merging code to contiv, but pull request is not created yet, so it should be ok i think)

PR #1288, opened 2019-04-16
Telemetry for VPP 19.04 (use stats API + fix CLI parsing)

ondrej-fabry:


plugins/telemetry/prometheus.go
```diff
@@ -248,10 +252,11 @@ func (p *Plugin) updatePrometheus() {
 	}
 
 	// Update memory
-	memoryInfo, err := p.handler.GetMemory()
+	/*memoryInfo, err := p.handler.GetMemory()
```
VladoLavor: Perhaps some info why it was left here commented.

ondrej-fabry: That was commented by accident. Fixed now.

PR #1294, opened 2019-04-24
unit test watch_vppcals

AndrejKilvady:
Signed-off-by: AndrejKilvady <andrej.kilvady@pantheon.tech>

### DIFF
--- a/plugins/vpp/ifplugin/vppcalls/vpp1901/watch_vppcalls.go
+++ b/plugins/vpp/ifplugin/vppcalls/vpp1901/watch_vppcalls.go
@@ -108,8 +108,6 @@ func (h *InterfaceVppHandler) WatchDHCPLeases(leasesCh chan<- *vppcalls.Lease) e
 					continue
 				}
 				lease := dhcpEvent.Lease
-				var hostMac net.HardwareAddr
-				copy(hostMac, lease.HostMac)
 				var hostAddr, routerAddr string
 				if uintToBool(lease.IsIPv6) {
 					hostAddr = fmt.Sprintf("%s/%d", net.IP(lease.HostAddress).To16().String(), uint32(lease.MaskWidth))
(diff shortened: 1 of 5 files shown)


plugins/vpp/ifplugin/vppcalls/vpp1901/watch_vppcalls.go
```diff
@@ -108,8 +108,7 @@ func (h *InterfaceVppHandler) WatchDHCPLeases(leasesCh chan<- *vppcalls.Lease) e
 					continue
 				}
 				lease := dhcpEvent.Lease
-				var hostMac net.HardwareAddr
-				copy(hostMac, lease.HostMac)
```
ondrej-fabry: Why remove this? Was this causing some issue for you? 

The reason why the `copy` function was used here, was because the [`net.HardwareAddr`](https://godoc.org/net#HardwareAddr) is defined as `[]byte`, slice of bytes, and slices in go are referenced by a pointer. This means that in this case if the `lease.HostMac` field in lease event was to be modified by some other goroutine it will change `hostMac` variable too. By using `copy` we copy the entire slice to a local variable, thus avoiding this situation.

Anyway.. this article probably explains it better than me: https://www.callicoder.com/golang-slices/

AndrejKilvady: I changed it. It's now OK?

ondrej-fabry: On travis you can see that it's failing with:

```     Expected
            <string>:   00
        to be equivalent to
            <string>: 10:10:20:20:30:30
```

You converted `[]byte` to string rather than using `String()` method of `net.HardwareAddr`. See the difference here: https://play.golang.org/p/Tc4aZ2UhPu3

plugins/vpp/ifplugin/vppcalls/vpp1901/watch_vppcalls_test.go
```diff
@@ -0,0 +1,151 @@
...
+	Expect(err).To(BeNil())
+
+	notifChan <- &interfaces.SwInterfaceEvent{
+		SwIfIndex:   1,
+		AdminUpDown: 1,
+		LinkUpDown:  1,
+		Deleted:     1,
+	}
+	var result *vppcalls.InterfaceEvent
+	checkIfState(1, result, eventsChan)
```
ondrej-fabry: Checking the expected results in `checkIfState` just creates unnecessary indirection and requires reader to jump up and down in code.

Try to contain tests inside single test function if possible. If you want to make it shorter you can use `Expect` to compare entire struct with values.

PR #1295, opened 2019-04-25
govppmux: allow to choose between shm & socket using a config option

milanlenco:
Some would want to have this configurable via the config file...


ondrej-fabry:
Also, to avoid confusion we should either deprecate `ShmPrefix` and use single option for both shared memory prefix and unix socket address.. or add unix socket address and choose which one of those to use in `NewVppAdapter`.

milanlenco:
Added option `BinApiSocketPath` to avoid the confusion.

ondrej-fabry:
Linter complaining.. requires `BinAPISocketPath`.

PR #1332, opened 2019-05-09
Support GSO flag for TAP interfaces in VPP 

ondrej-fabry:


plugins/vpp/ifplugin/descriptor/interface.go
```diff
@@ -688,21 +708,17 @@ func (d *InterfaceDescriptor) getMemifRingSize(memif *interfaces.MemifLink) uint
...
-		TxRingSize:     intf.GetTap().GetTxRingSize(),
-	}
-	if tapCfg.Version == 0 {
-		tapCfg.Version = 1
+func getTapConfig(intf *interfaces.Interface) (tapLink *interfaces.TapLink) {
+	tapLink = new(interfaces.TapLink)
+	proto.Merge(tapLink, intf.GetTap())
+
+	if tapLink.Version == 0 {
+		tapLink.Version = 1
```
ondrej-fabry: I actually kept defaulting of TAP version to `1`, since we still also support VPP 19.04 that still had version 1.

VladoLavor: In that case, you should allow configuring TAPv1 in vppcalls for vpp1904, because now it returns an error since the TAPv1 was deprecated before.

ondrej-fabry: I meant to write 19.01. VPP 19.04 dropped tapv1.

PR #1344, opened 2019-05-16
Refactoring Rx-placement and Rx-mode

milanlenco:
The original implementation:
- didn't allow to set rx-placement and rx-mode for multiple queues at once
- didn't work correctly with memifs - placement and mode cannot be set until memif has established connection, because only then the queues are created are ready to accept the configuration (therefore interface link state has to be watched and scheduler needs to be notified about changes)

As always I have also included small cleanup and bug fixing from inside kvscheduler...


api/models/vpp/interfaces/interface.proto
```diff
@@ -40,27 +40,30 @@ message Interface {
...
-        uint32 queue_id  = 2;
-        uint32 queue_id_valid  = 3;
+        uint32 queue = 1;
+        Type mode = 2;
+        bool default_mode = 3; // if enabled, <queue> will be ignored and the given
+                               // <mode> will be used for all queues without explicitly
+                               // selected Rx mode
     }
-    RxModeSettings rx_mode_settings = 10;
+    repeated RxMode rx_mode = 10;
```
ondrej-fabry: This would probably be better named as `rx_modes` to better represent that it's array.

api/models/vpp/interfaces/interface.proto
```diff
@@ -40,27 +40,30 @@ message Interface {
...
-        uint32 worker = 2;
-        bool is_main = 3;
+    message RxPlacement {
+        uint32 queue = 1;  // select from interval <0, number-of-queues)
+        uint32 worker = 2; // select from interval <0, number-of-workers)
+        bool main_thread = 3; // let the main thread to process the given queue
+                              //  - if enabled, value of <worker> is ignored
     }
-    RxPlacementSettings rx_placement_settings = 11;
+    repeated RxPlacement rx_placement = 11;
```
ondrej-fabry: And this `rx_placements`.

plugins/kvscheduler/internal/registry/registry_impl.go
```diff
@@ -111,8 +112,11 @@ func (reg *registry) GetDescriptorForKey(key string) *KVDescriptor {
 	var keyDescriptor *KVDescriptor
 	for _, descriptor := range reg.descriptors {
 		if descriptor.KeySelector(key) {
+			if keyDescriptor != nil {
```
ondrej-fabry: Was just thinking if this is actually not gonna make `GetDescriptorForKey` more expensive performance-wise. Can you run the perf tests with 20k to compare difference?

plugins/vpp/vppcallmock/vpp_ctx_mock.go
```diff
@@ -165,9 +165,7 @@ func (ctx *TestCtx) MockReplies(dataList []*HandleReplies) {
 				sendControlPing = dataMock.Ping
 				if len(dataMock.Messages) > 0 {
 					log.DefaultLogger().Infof(" MOCK HANDLER: mocking %d messages", len(dataMock.Messages))
-					for _, msg := range dataMock.Messages {
-						ctx.MockVpp.MockReply(msg)
-					}
+					ctx.MockVpp.MockReply(dataMock.Messages...)
```
ondrej-fabry: How come it was working until now when the `MockReply` creates additional reply for each time it's called? Or was it unused and was just unnoticed?

milanlenco: yeah, unused until now

plugins/kvscheduler/internal/registry/registry_impl.go
milanlenco: It varies quite a bit on my system, but there doesn't seem to measurable drop in the performance:
```
BEFORE:

level=info msg="Client #0 done, 20000 tunnels took 1m26.308s" loc="grpc-perf/main.go(302)" logger=grpc-stress-test-client
level=info msg="All clients done, took: 1m16.7757s" loc="grpc-perf/main.go(219)" logger=grpc-stress-test-client
level=info msg="All clients done, took: 1m15.096s" loc="grpc-perf/main.go(219)" logger=grpc-stress-test-client

AFTER:

level=info msg="All clients done, took: 1m5.0734s" loc="grpc-perf/main.go(219)" logger=grpc-stress-test-client
level=info msg="All clients done, took: 1m18.1164s" loc="grpc-perf/main.go(219)" logger=grpc-stress-test-client
level=info msg="All clients done, took: 1m12.1412s" loc="grpc-perf/main.go(219)" logger=grpc-stress-test-client
```

This lookup is cached (for the last 500 different keys), so effectively each distinct key goes through the cycle only once and the number of descriptors isn't that significant.

ondrej-fabry: Awesome, merging then.

PR #1357, opened 2019-05-27
Support for VPP 19.08-rc0

ondrej-fabry:
- [x] update GoVPP with support for field meta info (limit)
- [x] generate binapi for VPP 19.08
- [x] add vppcalls for VPP 19.08


plugins/vpp/natplugin/vppcalls/vpp1908/dump_nat_vppcalls.go
```diff
@@ -0,0 +1,578 @@
...
+	if flags&ba_nat.NAT_IS_TWICE_NAT != 0 {
+		natFlags.isTwiceNat = true
+	}
+	return natFlags
+}
+
+func addressTo4IP(address ba_nat.IP4Address) string {
+	ipAddr := make(net.IP, net.IPv4len)
+	copy(ipAddr[:], address[:])
+	if ipAddr.To4() == nil {
```
ondrej-fabry: This case will never happen. The `ipAddr` is allocated with length of IPv4 and first check in `To4()` returns the IP itself in that case, thus never returning `nil` here.

Also, since this function returns string type, which is immutable in Go, there is no need to copy anything and you can do the same by simply converting to `net.IP` and calling `String()` directly like this:

```go
net.IP(ip[:]).String()
```

plugins/vpp/natplugin/vppcalls/vpp1908/dump_nat_vppcalls_test.go
```diff
@@ -0,0 +1,379 @@
...
+	return ctx, natHandler, swIfIndexes, dhcpIndexes
+}
+func ipTo4Address(ipStr string) (addr bin_api.IP4Address) {
+	netIP := net.ParseIP(ipStr)
+	if ip4 := netIP.To4(); ip4 != nil {
+		var ip4Addr bin_api.IP4Address
+		copy(ip4Addr[:], netIP.To4())
+		addr = ip4Addr
+	} else {
+		return addr
```
ondrej-fabry: This case does nothing. Both statements `return` and `return addr` do same thing.

plugins/vpp/natplugin/vppcalls/vpp1908/dump_nat_vppcalls_test.go
```diff
@@ -0,0 +1,379 @@
...
+	swIfIndexes := ifaceidx.NewIfaceIndex(logrus.DefaultLogger(), "test-sw_if_indexes")
+	dhcpIndexes := idxmap_mem.NewNamedMapping(logrus.DefaultLogger(), "test-dhcp_indexes", nil)
+	natHandler := vpp1908.NewNatVppHandler(ctx.MockChannel, swIfIndexes, dhcpIndexes, log)
+	return ctx, natHandler, swIfIndexes, dhcpIndexes
+}
+func ipTo4Address(ipStr string) (addr bin_api.IP4Address) {
+	netIP := net.ParseIP(ipStr)
+	if ip4 := netIP.To4(); ip4 != nil {
+		var ip4Addr bin_api.IP4Address
+		copy(ip4Addr[:], netIP.To4())
```
ondrej-fabry: YOu can use `ip4` instead of calling `To4()` again here.

plugins/vpp/natplugin/vppcalls/vpp1908/dump_nat_vppcalls_test.go
```diff
@@ -0,0 +1,379 @@
...
+	log := logrus.NewLogger("test-log")
+	swIfIndexes := ifaceidx.NewIfaceIndex(logrus.DefaultLogger(), "test-sw_if_indexes")
+	dhcpIndexes := idxmap_mem.NewNamedMapping(logrus.DefaultLogger(), "test-dhcp_indexes", nil)
+	natHandler := vpp1908.NewNatVppHandler(ctx.MockChannel, swIfIndexes, dhcpIndexes, log)
+	return ctx, natHandler, swIfIndexes, dhcpIndexes
+}
+func ipTo4Address(ipStr string) (addr bin_api.IP4Address) {
+	netIP := net.ParseIP(ipStr)
+	if ip4 := netIP.To4(); ip4 != nil {
+		var ip4Addr bin_api.IP4Address
```
ondrej-fabry: No need for another `IP4Address` there is the `addr` already, you could copy directly into it.

plugins/govppmux/vppcalls/vpp1908/vpe_vppcalls.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/ligato/vpp-agent/plugins/vpp/binapi/vpp1908/memclnt"
+	"github.com/ligato/vpp-agent/plugins/vpp/binapi/vpp1908/vpe"
+)
+
+func init() {
+	var msgs []govppapi.Message
+	msgs = append(msgs, vpe.Messages...)
+	msgs = append(msgs, memclnt.Messages...)
+
+	vppcalls.Versions["vpp1904"] = vppcalls.HandlerVersion{
```
ondrej-fabry: Should be `vpp1908`.

PR #1360, opened 2019-05-29
Minor Code Fixes

matus-mrekaj:


pkg/models/encoding.go
```diff
@@ -113,7 +113,8 @@ func getItemID(pb proto.Message) (*api.Item_ID, error) {
 		return nil, fmt.Errorf("message %s is not registered as model", protoName)
 	}
 
-	name, err := model.nameFunc(pb)
+	//name, err := model.nameFunc(pb)
```
VladoLavor: Please remove the commented code.

VladoLavor:
Please do the PR into the 'dev' branch.  We do PRs to master only when releasing or when necessary for some reason (and in that case, it's good to double-commit to the `dev` as well).

PR #1364, opened 2019-06-03
Add config status publishing for config items

ondrej-fabry:


client/client_api.go
```diff
@@ -4,13 +4,18 @@ import (
 	"context"
 
 	"github.com/gogo/protobuf/proto"
+
 	api "github.com/ligato/vpp-agent/api/genericmanager"
 )
 
+type ModelInfo = api.ModelInfo
+
+type StateItem = api.StateItem
```
VladoLavor: Isn't it redundant creating aliases with the same name as original types?

examples/custom_model/main.go
```diff
@@ -109,12 +110,14 @@ func (p *ExamplePlugin) AfterInit() (err error) {
...
-
+		// remoteclient
+		c := remoteclient.NewClientGRPC(genericmanager.NewGenericManagerClient(p.conn))
 		demonstrateClient(c)
 
-		time.Sleep(time.Second * 3)
+		//time.Sleep(time.Second * 3)
+
+		// localclient
+		//demonstrateClient(client.LocalClient)
```
VladoLavor: You could remove all this clutter.

client/client_api.go
ondrej-fabry: These will be defined as new types later and the API won't have to change for the rest of the code.

examples/custom_model/main.go
ondrej-fabry: Not a clutter. This is actually how the example will work after some upcoming changes. 

This example will most likely ask user to pick client implementation used for configuring agent. This should demonstrate that the client API can be kept exactly same for all client implementations (local, remote grpc, remote kvdb...)

VladoLavor: Then I would comment it with a little explanation so it does not look like forgotten commented code, for example:
````
// choose client implementation (remote, local, ...)

// remoteclient
c := remoteclient.NewClientGRPC(genericmanager.NewGenericManagerClient(p.conn))
demonstrateClient(c)

// localclient
//demonstrateClient(client.LocalClient)	

...
````
plus `time.Sleep()` where needed

ondrej-fabry: The localclient is not fully supported. I will un-comment this and ask for user input to pick implementation when all are ready and fully implemented.

Issue #1365, opened 2019-06-04
vpp-agent does not configure rx-mode


miroslavKovacPantheon:
vpp version 1904
vpp-agent version v2.2.0-alpha-45-ge39b5b6" BuildDate="2019-05-31T01:28-07:00" CommitHash=e39b5b6578699d99b7268034fe0de1aeb27b7fde

rx-mode is not always set up. sometimes I see that memif is set to polling mode even if we set it to adaptive.. 
etcdctl output ->
```
/vnf-agent/vpp-unit-18/config/vpp/l2/v2/bridge-domain/L2BR18101
{"name": "L2BR18101", "flood": true, "unknown_unicast_flood": true, "forward": true, "learn": true, "mac_age": 5, "interfaces": [{"name": "loop181", "bridged_virtual_interface": true}, {"name": "memif181"}]}
/vnf-agent/vpp-unit-18/config/vpp/nat/v2/nat44-global
{"nat_interfaces": [{"name": "loop181", "is_inside": true}, {"name": "memif180"}], "address_pool": [{"address": "66.253.160.17"}]}
/vnf-agent/vpp-unit-18/config/vpp/v2/dhcp-proxy/IPv4
{"source_ip_address": "192.168.0.19", "servers": [{"ip_address": "66.253.163.250"}]}
/vnf-agent/vpp-unit-18/config/vpp/v2/interfaces/loop181
{"name": "loop181", "type": "SOFTWARE_LOOPBACK", "enabled": true, "ip_addresses": ["100.70.17.1/24"]}
/vnf-agent/vpp-unit-18/config/vpp/v2/interfaces/loop182
{"name": "loop182", "type": "SOFTWARE_LOOPBACK", "enabled": true, "ip_addresses": ["66.253.160.17/32"]}
/vnf-agent/vpp-unit-18/config/vpp/v2/interfaces/memif180
{"name": "memif180", "type": "MEMIF", "enabled": true, "ip_addresses": ["192.168.0.19/22"], "rx_modes": [{"mode": "ADAPTIVE", "default_mode": true}], "memif": {"socket_filename": "/run/memif/18.sock"}}
/vnf-agent/vpp-unit-18/config/vpp/v2/interfaces/memif181
{"name": "memif181", "type": "MEMIF", "enabled": true, "rx_modes": [{"mode": "ADAPTIVE", "default_mode": true}], "memif": {"id": 1, "socket_filename": "/run/memif/18.sock"}}
[...]

PR #1367, opened 2019-06-04
Add some (temporary) debug logs for interface state updates

milanlenco:
Added temporary logs to help with the issue https://github.com/ligato/vpp-agent/issues/1365
In the reported case it seems that ALL interface state notifications are lost somewhere along the way...


plugins/vpp/ifplugin/descriptor/link_state.go
```diff
@@ -130,6 +132,7 @@ func (w *LinkStateDescriptor) UpdateLinkState(ifaceState *interfaces.InterfaceNo
 				Value:    nil,
 				Metadata: nil,
 			})
+			delete(w.linkStates, ifaceName)
```
milanlenco: btw. found one bug here, but this is not related to the reported issue

PR #1370, opened 2019-06-05
fix ARM64 docker image build

stanislav-chlebec:
For building VPP was used gcc-7 which ended by failure on ARM64 platform
gcc-8 and g++-8 are installed and used instead to ensure successfull build
VPP is built with gcc-8 for both platform variants - x86_64 and aarch64 successfully

Signed-off-by: Stanislav Chlebec <stanislav.chlebec@pantheon.tech>


stanislav-chlebec:
I found out that there are the same errors , so I continue to inspect ....

PR #1384, opened 2019-06-11
vpp.punt: Add support for punt exceptions

ondrej-fabry:
- added support for punt exceptions in puntplugin
- fixed deleting of registered sockets after resync
- forcing recreation of registered sockets using prefix '!' in socket path
- SetPunt is not used anymore


plugins/vpp/puntplugin/puntplugin.go
```diff
@@ -83,10 +93,10 @@ func (p *PuntPlugin) Init() (err error) {
 		return err
 	}
 
-	// init and register IP punt redirect
-	p.ipRedirectDescriptor = descriptor.NewIPRedirectDescriptor(p.puntHandler, p.Log)
-	ipRedirectDescriptor := adapter.NewIPPuntRedirectDescriptor(p.ipRedirectDescriptor.GetDescriptor())
-	err = p.KVScheduler.RegisterKVDescriptor(ipRedirectDescriptor)
+	// init and register punt exception descriptor
+	p.puntExceptionDescriptor = descriptor.NewPuntExceptionDescriptor(p.puntHandler, p.Log)
```
VladoLavor: You can you shorten descriptor registration (as for the [ifplugin ](https://github.com/ligato/vpp-agent/blob/dev/plugins/vpp/ifplugin/ifplugin.go#L160) for example).

ondrej-fabry: Perhaps later with all other ones.

PR #1397, opened 2019-06-28
Add debug package for profiling/debugging

ondrej-fabry:


pkg/debug/debug.go
```diff
@@ -0,0 +1,99 @@
...
+	profilePath     = os.Getenv("DEBUG_PROFILE_PATH")
+	debugServerAddr = os.Getenv("DEBUG_SERVER_ADDR")
+)
+
+type Debug struct {
+	closer func()
+}
+
+func Start() interface {
+	Stop()
```
VladoLavor: Question: this is very interesting and I didn't know it is possible - but is there any advantage against just returning `*Debug`?

ondrej-fabry: To make it clear to the user that at that point only `Stop` should be called.

PR #1414, opened 2019-07-19
Fix in Telemetry plugin and improvements for integration tests

ondrej-fabry:


plugins/govppmux/vppcalls/vppcalls_api.go
```diff
@@ -15,6 +15,14 @@ type VersionInfo struct {
 	BuildDirectory string
 }
 
+// Release returns version in shortened format YY.MM that describes release.
+func (v VersionInfo) Release() string {
+	if len(v.Version) < 5 {
```
VladoLavor: What the number represents? Wouldn't it be better as a constant?

ondrej-fabry: Just as the comment is tryig to explain, it trims the VPP version from `v19.08-rc0~680-g456d2f9c2~b3246` to `19.08`.

VladoLavor: Agree in this case

PR #1415, opened 2019-07-22
Optimize interface state dump for specific interfaces

ondrej-fabry:
- define env var for printing debug logs about interface states
- extend interface state info in vppcalls for ifplugin
- return error when loading config fails
- minor improvements for VPP integration testing


plugins/vpp/ifplugin/vppcalls/vpp1908/dump_interface_vppcalls.go
```diff
@@ -161,6 +148,130 @@ func (h *InterfaceVppHandler) dumpInterfaces(ifIdxs ...uint32) (map[uint32]*vppc
 	return ifs, nil
 }
 
+func (h *InterfaceVppHandler) DumpInterfaces2(ifIdxs ...uint32) (map[uint32]*vppcalls.InterfaceDetails, error) {
```
VladoLavor: Did you forget to finish this one? The implementation looks the same as  `DumpInterfaces` except the index check and I did not find where is it called. If both of them were intended to remain here, please remove duplicated code and find some more suitable name, this does not look good.

ondrej-fabry:
> We are having more and more environment variables, we should also start documenting it.

If you refer to the `DEBUG_*` ones I think those are for very specific debug purpose and I would not clutter documentation with it.

PR #1417, opened 2019-07-23
Add option for skipping metrics in telemetry plugin

ondrej-fabry:
This allows user to skip some of the metrics in telemetry plugin using option in configuration file, e.g `skipped: [nodes]` for skipping node metrics


ondrej-fabry:
I would definitely keep both. We don't want to break anyone's setup where they currently disable the plugin with `disabled: true`.

PR #1420, opened 2019-07-25
Minor enhancements and code cleanups

ondrej-fabry:


plugins/linux/ifplugin/descriptor/interface_tap.go
```diff
@@ -67,8 +70,7 @@ func (d *InterfaceDescriptor) createTAPToVPP(nsCtx nslinuxcalls.NamespaceMgmtCtx
 	defer revert()
 
 	// rename from temporary host name to the request host name
-	d.ifHandler.RenameInterface(vppTapHostName, hostName)
```
ondrej-fabry: This might have returned error without logging it before.

Issue #1418, opened 2019-07-25
RX placement config causes KVScheduler to panic


ondrej-fabry:
KVScheduler panics in some cases during processing of transaction in which the RX placements gets updated to have the same queue ID multiple times. 

For example following change from this:

```sh
name:"memif2"
type:MEMIF 
enabled:true 
rx_modes:<mode:DEFAULT default_mode:true >
rx_placements:<main_thread:true >
memif:<
  master:true 
  id:1 
  rx_queues:4 
  tx_queues:4 
>  
```

to this:
```sh
name:"memif2"
type:MEMIF 
enabled:true 
rx_modes:<mode:DEFAULT default_mode:true > rx_modes:<queue:1 mode:POLLING > 
rx_placements:<queue:1 worker:1 > rx_placements:<queue:1 main_thread:true > 
memif:<
  master:true 
  id:1
  rx_queues:4 
  tx_queues:4 
>  
```

Here are 3 panics in different parts of code:

<details><summary><b>1. panic</b> (reported by Vinay)</summary><p>

```
+======================================================================================================================+
| Transaction #10                                                                                                RETRY |
+======================================================================================================================+
  * transaction arguments:
      - seq-num: 10
      - type: RETRY (for txn 6, attempt #1)
      - values:
          - key: config/vpp/v2/interfaces/memif2
            val: { name:"memif2" type:MEMIF enabled:true mtu:1500 rx_placements:<queue:1 worker:1 > rx_placements:<queue:2 worker:2 > memif:<master:true id:2 socket_filename:"/run/vpp/memif.sock" >  }
  * planned operations:
      1. DELETE [DERIVED RETRY RECREATE]:
          - key: vpp/interface/memif2/rx-placement/queue/2
          - value: { queue:2 worker:1  }
          - prev-error: failed to set rx-placement for queue 2 of the interface memif2: VPPApiError: Unimplemented (-9)
      2. CREATE [DERIVED RETRY RECREATE]:
          - key: vpp/interface/memif2/rx-placement/queue/2
          - value: { queue:2 worker:2  }

time="2019-07-23 22:01:38.72777" level=error msg="failed to set rx-placement for queue 2 of the interface memif2: VPPApiError: Unimplemented (-9)" loc="descriptor/rx_placement.go(98)" logger=vpp-ifplugin.rx-placement-descriptor

panic: runtime error: invalid memory address or nil pointer dereference
[...]

PR #1423, opened 2019-07-29
kvscheduler: keep node in the graph during recreate

milanlenco:
This patch fixes issues reported in #1418.

During recreate (Delete+Create) the node would be completely
removed from the graph by Delete and all the associated flags
(metadata) were therefore lost.
When the subsequent Create fails, the scheduler needs to access
the flags, which in this case are undefined, causing the scheduler
to panic by dereferencing nil pointer.
This patch ensures that the node and its flags are preserved
during re-creation.

An in-progress refactor of the scheduling algorithm
will approach value recreation in a much cleaner
way...


ondrej-fabry:
@milanlenco  Is there some unit test that could be added for this case?

milanlenco:
Added UT that would fail with panic without this patch.

PR #1425, opened 2019-07-30
Agentctl

ondrej-fabry:


cmd/agentctl/commands/import.go
```diff
@@ -43,73 +43,97 @@ var (
...
+    <keyN> <valueN>
+
+ Empty lines and lines starting with '#' are ignored.
+
+KEY FORMAT
+
+ Keys can be defined in two ways:
+
+    - full: 	/vnf-agent/vpp1/config/vpp/v2/interfaces/iface1
+    - short:	config/vpp/v2/interfaces/iface1
```
VladoLavor: I could not make the `input` command work properly with `short` key, perhaps it needs more detailed help. Using it as it returns `no model registered for key` error. Below is written that it needs `--label` flag, but where? Using it with import returns `Error: unknown flag: --label`. The `full` key works with no issues.

cmd/agentctl/commands/kvdb.go
```diff
@@ -0,0 +1,200 @@
...
+import (
+	"errors"
+	"fmt"
+
+	"github.com/spf13/cobra"
+)
+
+func NewKvdbCommand(cli *AgentCli) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "kvdb",
```
VladoLavor: This command returns `Failed to connect to Etcd: dial tcp: missing address` for all sub-commands. I do not see any flag where the address can be provided. Also, it should recognize default IP as other commands.

cmd/agentctl/commands/log.go
```diff
@@ -0,0 +1,143 @@
...
+	"os"
+	"strings"
+
+	"github.com/spf13/cobra"
+
+	"github.com/ligato/cn-infra/logging"
+	"github.com/ligato/vpp-agent/cmd/agentctl/utils"
+)
+
+func NewLogCommand(cli *AgentCli) *cobra.Command {
```
VladoLavor: Command `log set` could return an error when attempting to set the log level to a non-existing logger, or when using invalid log level.

cmd/agentctl/commands/import.go
VladoLavor: works with the latest commit, however, please update the flag in help to `--service-label `

cmd/agentctl/commands/generate.go
```diff
@@ -73,6 +75,7 @@ func runGenerate(cli *AgentCli, opts GenerateOptions) {
 	var err error
 
 	switch strings.ToLower(opts.Format) {
+
```
VladoLavor: Why blank lines?

PR #1428, opened 2019-07-31
WIP integration test ACL IP CRUD

stanislav-chlebec:


tests/integration/vpp/04_acl_test.go
```diff
@@ -0,0 +1,499 @@
...
+	}
+
+	_, err = h.AddACL(aclErr3Rules, "test4")
+	Expect(err).To(Not(BeNil()))
+	if err != nil {
+		t.Logf("adding acls failed: %v", err)
+	}
+
+	//ctx.MockVpp.MockReply(&acl_api.MacipACLAddReply{})
+	aclIdx, err = h.AddACL(aclIPrules, "test5")
```
stanislav-chlebec: Why in unit tests at this place is expected to be err not NIL

ondrej-fabry: https://github.com/ligato/vpp-agent/blob/dev/plugins/vpp/aclplugin/vppcalls/vpp1908/acl_vppcalls_test.go#L271-L277

This part is testing: invalid reply (wrong reply message) and error reploy (retval != 0). These test cases in this form are not relevant in the integration tests. However it would be good to test cases where error happens as well. For example trying to use invalid configuration and expect error from VPP.. etc.

tests/integration/vpp/040_acl_test.go
```diff
@@ -0,0 +1,614 @@
...
+	if errx != nil {
+		t.Fatalf("dumping acls failed: %v", errx)
+	}
+	t.Logf("%d acls dumped", len(acls))
+	t.Logf("acls dumped %v", acls)
+
+	aclIdx, err := h.AddACL(aclIPrules, "test0")
+	Expect(err).To(BeNil())
+	Expect(aclIdx).To(BeEquivalentTo(0))
+	if err != nil {
```
ondrej-fabry: There is no need to do `if err != nil` checks if you are using `Expect(err).To(BeNil())` from gomega package.

ondrej-fabry: https://onsi.github.io/gomega/#handling-errors

tests/integration/vpp/040_acl_test.go
```diff
@@ -0,0 +1,614 @@
...
+		SwIfIndex: ifIdx,
+	})
+
+	h := aclplugin_vppcalls.CompatibleACLVppHandler(ctx.vppBinapi, ifIndexes, logrus.NewLogger("test"))
+	if h == nil {
+		t.Fatalf("handler was not created")
+	}
+
+	acls, errx := h.DumpACL()
+	if errx != nil {
```
ondrej-fabry: Let's be consistent and use `Expect(...)...` for all of these from this point below.

tests/integration/vpp/040_acl_test.go
```diff
@@ -0,0 +1,614 @@
...
+		t.Fatalf("dumping acls failed: %v", errx)
+	}
+	t.Logf("%d acls dumped", len(acls))
+	t.Logf("acls dumped %v", acls)
+
```
ondrej-fabry: There should be some check here that would check number of dumped acls to be 0.

plugins/vpp/aclplugin/vppcalls/acl_vppcalls.go
```diff
@@ -82,13 +82,13 @@ type ACLVppAPI interface {
 	AddACLToInterfaceAsIngress(aclIndex uint32, ifName string) error
 	// AddACLToInterfaceAsEgress adds ACL (L3/L4) to single interface as egress.
 	AddACLToInterfaceAsEgress(aclIndex uint32, ifName string) error
-	// AddACLToInterfaceAsIngress deletes ACL (L3/L4) from single interface as ingress.
+	//DeleteACLFromInterfaceAsIngress deletes ACL (L3/L4) from single interface as ingress.
```
ondrej-fabry: Comments should start with space.

tests/integration/vpp/040_acl_test.go
```diff
@@ -0,0 +1,993 @@
...
+	aclplugin_vppcalls "github.com/ligato/vpp-agent/plugins/vpp/aclplugin/vppcalls"
+	_ "github.com/ligato/vpp-agent/plugins/vpp/ifplugin"
+	"github.com/ligato/vpp-agent/plugins/vpp/ifplugin/ifaceidx"
+	ifplugin_vppcalls "github.com/ligato/vpp-agent/plugins/vpp/ifplugin/vppcalls"
+	//	"encoding/json"
+)
+
+var aclNoRules []*acl.ACL_Rule
+
+var aclErr1Rules = []*acl.ACL_Rule{
```
ondrej-fabry: It's not that safe to define global variables like this and also takes a lot of space. 

It would be better to create a helper function that returns ACL rule, something like:

```go
func newAclIpRule(src, dst string) *acl.ACL_Rule {
	return &acl.ACL_Rule{
		Action: acl.ACL_Rule_PERMIT,
		IpRule: &acl.ACL_Rule_IpRule{
			Ip: &acl.ACL_Rule_IpRule_Ip{
				SourceNetwork:      src,
				DestinationNetwork: dst,
			},
		},
	}
}

func Test(...) {
	aclRules := []*acl.ACL_Rule{
		newAclIpRule(".0.", "10.20.0.0/24"),
	}
}
```

stanislav-chlebec: so far partly done on next commit
still to need to do for denyICMP denyTCP denyUDP and such...

tests/integration/vpp/040_acl_test.go
```diff
@@ -0,0 +1,906 @@
...
+			SourceAddressPrefix:  srcPrefix,
+			SourceMacAddress:     srcMacAddr,
+			SourceMacAddressMask: srcMacMask,
+		},
+	}
+}
+
+var aclNoRules []*acl.ACL_Rule
+
+var aclErr1Rules = []*acl.ACL_Rule{
```
ondrej-fabry: You plan to remove this right? The idea was to remove the global variables altogether and just call functions from tests.

tests/integration/vpp/040_acl_test.go
```diff
@@ -0,0 +1,906 @@
...
+var aclErr2Rules = []*acl.ACL_Rule{
+	newAclIpRule("192.168.1.1/32", ".0."),
+}
+
+var aclErr3Rules = []*acl.ACL_Rule{
+	newAclIpRule("192.168.1.1/32", "dead::1/64"),
+}
+
+var aclErr4Rules = []*acl.ACL_Rule{
+	newAclMacIpRule(true, "192.168.0.1", uint32(16), "", "ff:ff:ff:ff:00:00"),
```
ondrej-fabry: No need to convert to `uint32` you can simply use `16` here.

tests/integration/vpp/040_acl_test.go
```diff
@@ -0,0 +1,906 @@
...
+
+	h := aclplugin_vppcalls.CompatibleACLVppHandler(ctx.vppBinapi, ifIndexes, logrus.NewLogger("test"))
+	if h == nil {
+		t.Fatalf("handler was not created")
+	}
+
+	acls, errx := h.DumpMACIPACL()
+	Expect(errx).To(BeNil())
+	aclCnt := len(acls)
+	Expect(aclCnt).Should(Equal(0))
```
ondrej-fabry: You can use [`HaveLen(INT)`](https://onsi.github.io/gomega/#havelencount-int) matcher instead of this.

```go
Expect(acls).Should(HaveLen(0))
// in this case even
Expect(acls).Should(BeEmpty())
```

tests/integration/vpp/040_acl_test.go
```diff
@@ -0,0 +1,906 @@
...
+			t.Logf("%v", item.ACL.Interfaces)
+			for _, intf := range item.ACL.Interfaces.Ingress {
+				if intf == ifName {
+					isForInterface = true
+					break
+				}
+			}
+		}
+	}
+	if !isPresent {
```
ondrej-fabry: You can simplify this check using [`Annotated Assertions`](https://onsi.github.io/gomega/#annotating-assertions), like this:

```go
Expect(isPresent).To(BeFalse(), "Configured IP should be present")
```

tests/integration/vpp/040_acl_test.go
stanislav-chlebec: Yes, I admit that I did not understand at first.
I eliminated all variables related to ACL_rule.

tests/integration/vpp/040_acl_test.go
stanislav-chlebec: Thanks for advice

tests/integration/vpp/040_acl_test.go
stanislav-chlebec: Again thanks for advice

PR #1431, opened 2019-08-06
extend linux plugin

matus-mrekaj:


milanlenco:
So far looks good, but it will need couple more things to support resync:
- [Here](https://github.com/ligato/vpp-agent/blob/2b704f9c771e1a1f6256b76b110f372dae08f3b0/plugins/linux/ifplugin/descriptor/interface.go#L339) in Create you will need to configure alias for the interface to mark it as agent-configured and store its logical name to be able to implement the Retrieve function (i.e. function that reads the state of data-plane and returns the actual config, which is then compared with desired config to determine the set of operations to execute to get those two in-sync). For example, for TAPs (which also already exist in Create from the linux plugin point of view) this is done [here](https://github.com/ligato/vpp-agent/blob/master/plugins/linux/ifplugin/descriptor/interface_tap.go#L45-L52). I would suggest alias like: _agentPrefix + "/existing/" + logical-interface-name_
- Delete should then remove the alias, like it is done [here](https://github.com/ligato/vpp-agent/blob/master/plugins/linux/ifplugin/descriptor/interface_tap.go#L133-L142) (it will be probably better to move all the EXISTING-related stuff to interface_existing.go so that the main descriptor file contains only the code shared by all interface types)
- [here](https://github.com/ligato/vpp-agent/blob/master/plugins/linux/ifplugin/descriptor/interface.go#L863) in Retrieve the first if branch should check the alias if there is the "/existing/" substring before checking if it is TAP or VETH, and return it as EXISTING interface (with the name parsed from the alias) if it is the case.

With that the resync should work - to test that you can:
1. start agent and put some configuration into etcd (with EXISTING interface)
2. turn off the agent and change configured EXISTING interface(s) via Linux CLI (e.g. remove/add IP address)
3. (re)start the agent - the first transaction should re-read etcd content and get the network configuration in-sync - i.e. it should revert changes done in 2.

Similarly you can change the desired config in step 2. by changing the etcd content, but keep the network config (actual state) untouched, and then in step 3 the agent should reflect those changes.

matus-mrekaj:
I think the idea of type EXISTING should be just to configure interfaces that already existed before vpp-agent start(interfaces created by something other than the agent) or in the case that the interface is not present yet, configure it later when it is created.
In the case of `resync` it should be handled by the [watcher](https://github.com/ligato/vpp-agent/blob/dev/plugins/linux/ifplugin/descriptor/watcher.go#L109-L124) descriptor, which should take care of this [dependency](https://github.com/ligato/vpp-agent/blob/2b704f9c771e1a1f6256b76b110f372dae08f3b0/plugins/linux/ifplugin/descriptor/interface.go#L666-L671), since we expect the interface to be in the same namespace as the agent. If the interface is not present when the agent starts, then it will wait for the [SB notification](https://github.com/ligato/vpp-agent/blob/dev/plugins/linux/ifplugin/descriptor/watcher.go#L254-L258)

In the case of DHCP-CNF let's say we have a [stub network](https://github.com/contiv/vpp/blob/master/k8s/examples/sfc/cnfs-linux.yaml#L9), and the dhcp server should listen on that interface, the interface must have a ip address assigned,I think In this case the agent should be able to configure the interface.
(I'm not quite sure how the vswitch agent works, or if it can configure the interface)

milanlenco:
The problem we are facing is that with the stub network the vswitch agent will create the interface, but it will not assign any IP address to it. If the agent from DHCP CNF does assign and IP address, but then later the vswitch agent restarts, it would run the resync and remove the IP address (since by the vswitch's NB configuration there should be no IP address) - i.e. I can't figure out how to solve the multiple-agents configuring the same interface scenario.
So my idea was to use a non-stub network with IP addresses (other network types to be implemented into Contiv) and let the vswitch to configure the IP address. The question is though, does dhcp depends only on interface being created, or does IP address already has to be assigned when dhcp server is started? Because this [dependency](https://github.com/ligato/vpp-agent/blob/2b704f9c771e1a1f6256b76b110f372dae08f3b0/plugins/linux/ifplugin/descriptor/interface.go#L666-L671) does not cover interface having an IP address, so we would need to also watch for IP address add/del events into the watcher.

matus-mrekaj:
The dhcp depends on the interface to be up and have an ip address assigned otherwise it will not listen on that interface 
So if we set a config, for example with ETCD, we get this error:
 `2019-07-17 08:28:10.721 WARN  [kea-dhcp4.dhcpsrv/9] DHCPSRV_OPEN_SOCKET_FAIL failed to open socket: the interface tap1 has no usable IPv4 addresses configured
2019-07-17 08:28:10.721 WARN  [kea-dhcp4.dhcpsrv/9] DHCPSRV_NO_SOCKETS_OPEN no interface configured to listen to DHCP traffic`)
So to bypass this error the DHCP-CNF plugin has a dependency that wait until that interface is up and has a ip address assigned something like [this](https://github.com/ligato/vpp-agent/blob/2b704f9c771e1a1f6256b76b110f372dae08f3b0/plugins/linux/l3plugin/descriptor/arp.go#L233-L249)
Otherwise the dhcp-server will not accept the config.

milanlenco:
Regarding vswitch restart - the dependency on the interface would be also lost and regained which would cause the delete+create sequence also for your dhcp object, i.e. the dhcp server would get restarted.

Regarding the resync: I had a talk about the issues and possible solutions with @rastislavszabo. We decided that for now we would go with the solution based on the EXISTING interface type - i.e. the interface would be created by vswitch, but IP address(es) and other parameters would be set by the agent inside CNF. Maybe in the future we would add option to tell the CNF that the IP address will be set by the vswitch and the CNF should just wait for it to appear. For now I would also suggest to disregard the resync-related issues since we do not need to support the restart scenarios for the planned demo. I will therefore merge this PR so that you are no longer blocked by this and can finalize the CNF implementation. In the meantime I hope to find some spare time that I would use to test some ideas that I have on how to approach the resync in multi-agent scenarios.

PR #1437, opened 2019-08-19
Distinguish localclient from other kvdb data sources

milanlenco:
Signed-off-by: Milan Lenco <milenco@cisco.com>

### DIFF
--- a/client/local_client.go
+++ b/client/local_client.go
@@ -24,6 +24,7 @@ import (
 
 	api "github.com/ligato/vpp-agent/api/genericmanager"
 	"github.com/ligato/vpp-agent/pkg/models"
+	orch "github.com/ligato/vpp-agent/plugins/orchestrator"
 )
 
 // LocalClient is global client for direct local access.
@@ -57,7 +58,9 @@ func (c *client) ResyncConfig(items ...proto.Message) error {
 		txn.Put(key, item)
 	}
 
-	return txn.Commit(context.Background())
+	ctx := context.Background()
+	ctx = orch.DataSrcContext(ctx, "localclient")
+	return txn.Commit(ctx)
 }
 
 func (c *client) GetConfig(dsts ...interface{}) error {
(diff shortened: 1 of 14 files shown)


client/local_client.go
```diff
@@ -57,7 +58,12 @@ func (c *client) ResyncConfig(items ...proto.Message) error {
 		txn.Put(key, item)
 	}
 
-	return txn.Commit(context.Background())
+	ctx := context.Background()
+	_, withDataSrc := orch.DataSrcFromContext(ctx)
```
ondrej-fabry: Will this even return `withDataSrc=true` if we use `context.Background()` here?

examples/localclient_with_etcd/main.go
```diff
@@ -0,0 +1,133 @@
...
+
+// Close cleans up the resources.
+func (p *ExamplePlugin) Close() error {
+	return nil
+}
+
+func resyncLocalClient() {
+	fmt.Println("=== RESYNC WITH MEMIF ===")
+
+	/*
```
ondrej-fabry: This needs info comment for users to explain why it's commented out.

Or it could be uncommented and placed in separate function like `resyncLocalClientLegacy`?

examples/localclient_with_etcd/main.go
```diff
@@ -0,0 +1,133 @@
...
+	)
+	if err := a.Run(); err != nil {
+		log.Fatal(err)
+	}
+}
+
+// ExamplePlugin is the main plugin which
+// handles resync and changes in this example.
+type ExamplePlugin struct {
+	KVScheduler  *kvs.Scheduler
```
ondrej-fabry: Do we need to have `KVScheduler` specifically listed, even if it's not used by the `ExamplePlugin`? The orchestrator plugin contains default KVScheduler instance already and is listed already by the `ExamplePlugin`.

client/local_client.go
milanlenco: if condition added by mistake - removed

examples/localclient_with_etcd/main.go
milanlenco: added another function for the legacy client and described briefly in the comments

examples/localclient_with_etcd/main.go
milanlenco: another mistake (the example was copy&pasted from another one which actually does something with the scheduler) - fixed

PR #1444, opened 2019-08-21
Linux ifplugin improvements + VPP ifplugin bug fix

milanlenco:
Linux ifplugin improvements:
- IP addresses are manipulated as derived values (like the VPP ifplugin does) - so, for example, interface is created even if some IP address is invalid / fails to be assigned
- implemented Retrieve for EXISTING interfaces (resync now works correctly)
- added "link_only" parameter into the model - if enabled, the linux plugin will only create the interface and let something else to manage addresses (could be another agent which manipulates with the interface as EXISTING, or admin managing addresses via CLI) - i.e. resync will not touch MAC and IP addresses

VPP ifplugin bug fix:
- derived IP addresses obtained from DHCP were excluded from the key space of InterfaceAddressDescriptor, because otherwise the descriptor would try to add the IP address again
- the derived keys for IP addresses are now:
    a) vpp/interface/{iface}/address/static/{address}: derived from interfaces, managed by InterfaceAddressDescriptor
    b) vpp/interface/{iface}/address/from-dhcp/{address} : derived from DHCP lease as PROPERTY (i.e. without descriptor, but can be referenced in dependencies)
- to depend on interface to have any IP address, use this key prefix: "vpp/interface/{iface}/address/"
- to depend on interface to have assigned at least one static IP address, use this key prefix: "vpp/interface/{iface}/address/static"
- to depend on interface to have assigned at least one IP address from DHCP, use this key prefix: "vpp/interface/{iface}/address/from-dhcp"


ondrej-fabry:
It seems to be that it would be quite useful to have integration tests for linux related config as well.

milanlenco:
Agree with the integration tests - although, for Linux the linuxcalls are very simple and not so interesting for testing. The descriptors, on the other hand, are much more complex (for example this PR makes changes in descriptors, but not in linuxcalls). I will try to prepare at least one IT with Kvscheduler and Linux descriptors initialized and available for testing (to have something like ctx := setupLinux(t)). Or maybe we could create another directory under tests/intergration, such as "kvscheduler", where we include the descriptors+kvscheduler for both VPP and Linux?

Issue #1450, opened 2019-08-28
IfPlugin publishes notifications only if the interface has state InterfaceState_DELETED


denis-tingajkin:
@ondrej-fabry Hello, 
We faced a strange problem related to vpp-agent notifications.
`Expected`: all notifications related to interface will publish.
`Actual`:  We get notifications only on interface deleted.

Question: Is actual behavior ok?

Also, I provided a potential naive solution. (See PR below.)

## Issue from nsm-repo
https://github.com/networkservicemesh/networkservicemesh/issues/1512

PR #1451, opened 2019-08-28
[WIP] [PoC] publish all notifications

denis-tingajkin:
## Motivation
https://github.com/ligato/vpp-agent/issues/1450

### DIFF
--- a/plugins/vpp/ifplugin/publish_state.go
+++ b/plugins/vpp/ifplugin/publish_state.go
@@ -6,10 +6,9 @@ import (
 	"github.com/ligato/cn-infra/datasync"
 	"github.com/ligato/cn-infra/health/statuscheck"
 	"github.com/ligato/cn-infra/health/statuscheck/model/status"
-	"github.com/pkg/errors"
-
 	"github.com/ligato/vpp-agent/api/models/vpp"
 	interfaces "github.com/ligato/vpp-agent/api/models/vpp/interfaces"
+	"github.com/pkg/errors"
 )
 
 // watchStatusEvents watches for resync event of interface state data.
@@ -140,11 +139,11 @@ func (p *IfPlugin) publishIfStateEvents() {
 					p.Log.Debugf("Updating link state: %+v", ifState)
 				}
 				p.linkStateDescriptor.UpdateLinkState(ifState)
-				if p.PushNotification != nil {
-					p.PushNotification(&vpp.Notification{
-						Interface: ifState,
-					})
-				}
+			}
+			if p.PushNotification != nil {
+				p.PushNotification(&vpp.Notification{
+					Interface: ifState,
+				})
 			}
 
 			p.publishLock.Unlock()


milanlenco:
Hi @denis-tingajkin,

could you re-test with this PR applied: https://github.com/ligato/vpp-agent/pull/1453 ?
This fix should give you all the interface UP/DOWN events. But note that notifications about packet/byte counter changes are intentionally not sent through GRPC since they are very frequent and not suitable for that transport we think.
We actually prefer to publish the counters using e.g. Prometheus. This is provided by the Telemetry plugin - for inspiration on how to deploy agent with Prometheus and display stats via Grafana, see https://github.com/ligato/vpp-agent/wiki/Telemetry.
Also, the interface plugin still uses the old and slow binary API based access to interface statistics which slows down the VPP main thread, whereas Telemetry plugins reads stats from a shared memory without blocking VPP.

edwarnicke:
> Hi @denis-tingajkin,
> 
> could you re-test with this PR applied: #1453 ?
> This fix should give you all the interface UP/DOWN events. But note that notifications about packet/byte counter changes are intentionally not sent through GRPC since they are very frequent and not suitable for that transport we think.

Would it be possible to get them at a requested interval?

> We actually prefer to publish the counters using e.g. Prometheus. This is provided by the Telemetry plugin - for inspiration on how to deploy agent with Prometheus and display stats via Grafana, see https://github.com/ligato/vpp-agent/wiki/Telemetry.

That's a good thing to have, but doesn't meet our need.  We don't need Ligato Prometheus stats, we need NSM Connection metrics... in other words... we need you to give them to us so we can correlate them correctly with other metadata to report them to other systems.  I get that you can't update us everytime you get updated (that's a lot of events).  But we would want to ask for them at some time interval and get updates at that interval.

> Also, the interface plugin still uses the old and slow binary API based access to interface statistics which slows down the VPP main thread, whereas Telemetry plugins reads stats from a shared memory without blocking VPP.

Again... we need the data directly, via GRPC, not Prometheus.

ondrej-fabry:
I went over the requirements with Ed today and planning to provide some simple way to consume metrics using custom time periods.

BoutheinaDab:
@ondrej-fabry  @edwarnicke Want to know please if vpp-agent is able know to provide interface metrics at time periods?

ondrej-fabry:
@BoutheinaDab check out #1495

denis-tingajkin:
@ondrej-fabry we found that StatsPollerService is part of telemetry plugin. We don't want to use the telemetry plugin on NSM side. Can we implement StatsPollerService without a telemetry plugin?

ondrej-fabry:
@denis-tingajkin check out #1532, it adds `PrometheusDisabled` option to disable collection and export for Prometheus, which is what you don't use anyway.

denis-tingajkin:
@ondrej-fabry thanks, I tried to use stats poller service and it is not working for me. Please take a look https://github.com/ligato/vpp-agent/issues/1538.

denis-tingajkin:
[StatsPollerService](https://github.com/ligato/vpp-agent/pull/1495) has been implemented. This PR can be closed.

PR #1463, opened 2019-09-11
Fix afpacket CRUD tests

samelias:
- fix "show hardware" command
- rework "Get Interface Presence" keyword to improve error messages

Signed-off-by: samuel.elias <samelias@cisco.com>


samelias:
This seems to fix l2xconnect CRUD as well

PR #1468, opened 2019-09-12
Add e2e test for SPAN feature

rewenset:


tests/e2e/030_span_test.go
```diff
@@ -0,0 +1,186 @@
...
+	)
+
+	Expect(ctx.getValueState(spanRx)).To(
+		Equal(kvs.ValueState_PENDING),
+		"SPAN is not in a `PENDING` state, but `InterfaceTo` is not ready",
+	)
+
+	// Check `show int span` output
+	var stdout bytes.Buffer
+	cmd := exec.Command("vppctl", "show", "int", "span")
```
milanlenco: Perhaps add helper function `execVppctl(command string) (output string, err error)` (or more generally `execCommand` to make it usable even beyond vppctl) to `e2e_test.go`. The
```
var stdout bytes.Buffer
exec.Command(...)
cmd.Stdout = &stdout
...
```
sequence is repeated twice in this file, also occurs [here](https://github.com/ligato/vpp-agent/blob/dev/tests/e2e/e2e_test.go#L249-L254) and most likely will be needed in future tests.

tests/e2e/e2e_test.go
```diff
@@ -205,6 +206,18 @@ func (ctx *testCtx) agentInSync() bool {
 	return true
 }
 
+// execVppctl returns output from vppctl for given action and arguments.
+func (ctx *testCtx) execVppctl(action string, args ...string) (*bytes.Buffer, error) {
```
milanlenco: I think it would be more practical to return output as string, it most cases it would be followed by `stdout.String()` anyway.

PR #1469, opened 2019-09-16
Fixed missed interface dependency for SRv6 L2 steering

fgschwan:
Signed-off-by: Filip Gschwandtner <filip.gschwandtner@pantheon.tech>

### DIFF
--- a/plugins/vpp/srplugin/descriptor/steering.go
+++ b/plugins/vpp/srplugin/descriptor/steering.go
@@ -217,12 +217,12 @@ func (d *SteeringDescriptor) Dependencies(key string, steering *srv6.Steering) (
 		}
 	}
 	// Interface dependency
-	if t, isL2Traffic := steering.Traffic.(*srv6.Steering_L2Traffic_); isL2Traffic {
-		intfaceName := strings.TrimSpace(t.L2Traffic.InterfaceName)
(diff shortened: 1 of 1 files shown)


plugins/vpp/srplugin/descriptor/steering.go
```diff
@@ -215,5 +216,15 @@ func (d *SteeringDescriptor) Dependencies(key string, steering *srv6.Steering) (
 			})
 		}
 	}
+	// Interface dependency
+	if t, isL2Traffic := steering.Traffic.(*srv6.Steering_L2Traffic_); isL2Traffic {
```
ondrej-fabry: There is no need to do this boilerplate with oneof fields. The generated code for `Steering` already contains getter method `GetL2Traffic()` that also includes extra safety check to prevent panics on nil data. Your solution can panic for at least two cases..

https://github.com/ligato/vpp-agent/blob/8cdc8da9b3b5aab68d6897487a665eacf54248a4/api/models/vpp/srv6/srv6.pb.go#L1086-L1091

fgschwan: fixed in https://github.com/ligato/vpp-agent/pull/1469/commits/18ff2974675fe04c2776938e493a7b64a8e43ab3

PR #1472, opened 2019-09-17
Remove VAT from ACL CRUD tests and refactor

samelias:
- replace hardcoded expected data with a template
- cleanup ACL test suite file
- fix ACL ICMP tests testing ACL UDP instead
- move ACL libraries and resources to subdirectories

Signed-off-by: samuel.elias <samelias@cisco.com>


tests/robot/libraries/acl/acl_etcd.robot
```diff
@@ -0,0 +1,133 @@
+[Documentation]     ACL-related keywords for working with ETCD
```
VladoLavor: In this file are some commented lines which could be cleaned up if not needed

Issue #1465, opened 2019-09-12
Investigate failing e2e tests for VPP 20.01


ondrej-fabry:
Some e2e tests are failing for VPP 20.01 on Travis.

First failing build on Travis here: https://travis-ci.com/ligato/vpp-agent/builds/126617460

Failing test cases:
- [ ] [`--- FAIL: TestLinkOnly (13.13s)`](https://travis-ci.com/ligato/vpp-agent/jobs/232707974#L3293)
- [ ] [`--- FAIL: TestIPWithNeighGW (17.24s)`](https://travis-ci.com/ligato/vpp-agent/jobs/232707974#L3650)
- [ ] [`--- FAIL: TestIPWithNonLocalGW (17.24s)`](https://travis-ci.com/ligato/vpp-agent/jobs/232707974#L4016)
- [ ] [`--- FAIL: TestVPPRoutesWithNetalloc (18.25s)`](https://travis-ci.com/ligato/vpp-agent/jobs/232707974#L4341)

rewenset:
seems like the issue is with `flags`
for example, to fix `TestIPWithNeighGW` you should change converting function, by adding `3` to it.
```go
# file: plugins/vpp/ifplugin/vppcalls/vpp2001/dump_interface_vppcalls.go
func ifAPIFlagState(flags vpp_ifs.IfStatusFlags) uint32 {
	if flags == vpp_ifs.IF_STATUS_API_FLAG_ADMIN_UP {
		return 1
	} else if flags == vpp_ifs.IF_STATUS_API_FLAG_LINK_UP {
		return 2
	} else if flags == 3 {
		return 3
	}
	return 0
}
```
Why `3`? Because it is what returned when interface is `enabled`.
```
loop0: 3 
```
But if `enabled = false` then `IF_STATUS_API_FLAG_LINK_UP` returned.
```
tap0: IF_STATUS_API_FLAG_LINK_UP 
```
And this is the reason, I think, behind `DownstreamResync` contains not only NOOP, but UPDATE too. With that `tap0` example:
```
+======================================================================================================================+
| #3 - NB Transaction                                                                                 DownstreamResync |
+======================================================================================================================+
  * transaction arguments:
      - seqNum: 3
      - type: NB Transaction, SB Sync

o----------------------------------------------------------------------------------------------------------------------o
  * executed operations (2019-09-25 12:08:02.442 +0000 UTC -> 2019-09-25 12:08:02.456 +0000 UTC, dur: 13ms):
[...]

PR #1479, opened 2019-09-25
fix(vppcalls): Mapping flag for interface states

rewenset:
Issue #1465

### DIFF
--- a/plugins/vpp/ifplugin/vppcalls/vpp2001/dump_interface_vppcalls.go
+++ b/plugins/vpp/ifplugin/vppcalls/vpp2001/dump_interface_vppcalls.go
@@ -105,7 +105,7 @@ func (h *InterfaceVppHandler) dumpInterfaces(ifIdxs ...uint32) (map[uint32]*vppc
 				Name: strings.TrimRight(ifDetails.Tag, "\x00"),
 				// the type may be amended later by further dumps
 				Type:        guessInterfaceType(ifaceName),
-				Enabled:     ifAPIFlagState(ifDetails.Flags) > 0,
+				Enabled:     isAdminStateUp(ifDetails.Flags),
 				PhysAddress: net.HardwareAddr(ifDetails.L2Address[:]).String(),
 				Mtu:         getMtu(ifDetails.LinkMtu),
 			},
@@ -114,8 +114,8 @@ func (h *InterfaceVppHandler) dumpInterfaces(ifIdxs ...uint32) (map[uint32]*vppc
 				SupSwIfIndex:   ifDetails.SupSwIfIndex,
 				L2Address:      physAddr,
 				InternalName:   ifaceName,
-				IsAdminStateUp: ifAPIFlagState(ifDetails.Flags) > 0,
-				IsLinkStateUp:  ifAPIFlagState(ifDetails.Flags) > 1,
+				IsAdminStateUp: isAdminStateUp(ifDetails.Flags),
+				IsLinkStateUp:  isLinkStateUp(ifDetails.Flags),
 				LinkDuplex:     uint32(ifDetails.LinkDuplex),
 				LinkMTU:        ifDetails.LinkMtu,
 				LinkSpeed:      ifDetails.LinkSpeed,
@@ -391,8 +391,8 @@ func (h *InterfaceVppHandler) DumpInterfaceStates(ifIdxs ...uint32) (map[uint32]
 				SwIfIndex:    uint32(ifDetails.SwIfIndex),
 				InternalName: strings.TrimRight(ifDetails.InterfaceName, "\x00"),
 				PhysAddress:  physAddr,
-				AdminState:   toInterfaceStatus(ifDetails.Flags),
-				LinkState:    toInterfaceStatus(ifDetails.Flags),
+				AdminState:   adminStateToInterfaceStatus(ifDetails.Flags),
+				LinkState:    linkStateToInterfaceStatus(ifDetails.Flags),
 				LinkDuplex:   toLinkDuplex(ifDetails.LinkDuplex),
 				LinkSpeed:    toLinkSpeed(ifDetails.LinkSpeed),
 				LinkMTU:      ifDetails.LinkMtu,
@@ -402,19 +402,6 @@ func (h *InterfaceVppHandler) DumpInterfaceStates(ifIdxs ...uint32) (map[uint32]
 	}
 
 	return ifStates, nil
-}
-
-func toInterfaceStatus(flags vpp_ifs.IfStatusFlags) ifs.InterfaceState_Status {
-	switch flags {
-	case 0:
-		return ifs.InterfaceState_DOWN
-	case 1:
-		return ifs.InterfaceState_UP // admin state
-	case 2:
-		return ifs.InterfaceState_UP // link state
-	default:
-		return ifs.InterfaceState_UNKNOWN_STATUS
-	}
 }
 
 func toLinkDuplex(duplex vpp_ifs.LinkDuplex) ifs.InterfaceState_Duplex {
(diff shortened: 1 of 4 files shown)


plugins/vpp/ifplugin/vppcalls/vpp2001/dump_interface_vppcalls.go
```diff
@@ -1151,7 +1151,9 @@ func ifAPIFlagState(flags vpp_ifs.IfStatusFlags) uint32 {
 	if flags == vpp_ifs.IF_STATUS_API_FLAG_ADMIN_UP {
 		return 1
 	} else if flags == vpp_ifs.IF_STATUS_API_FLAG_LINK_UP {
-		return 2
+		return 0
+	} else if flags == 3 {
+		return 3
```
milanlenco: It is probably a VPP bug that the "value 3" is not defined in the binary APIs. If it has not been changed by mistake (try to find the commit that introduced the change), please create a Jira issue on fd.io wiki. 
Also not sure if returning 0 for IF_STATUS_API_FLAG_LINK_UP is intended?

milanlenco:
> We should test interface link state somehow too. @milanlenco what do you think?

That would be great. Not sure though if it can be done in the integration tests or if it has to be part of e2e tests.

rewenset:
Yes, currently  link state will be wrong:
https://github.com/ligato/vpp-agent/blob/10336c576fae527496c070da4793fe86118d8b4f/plugins/vpp/ifplugin/vppcalls/vpp2001/dump_interface_vppcalls.go#L118
https://github.com/ligato/vpp-agent/blob/10336c576fae527496c070da4793fe86118d8b4f/plugins/vpp/ifplugin/vppcalls/vpp2001/dump_interface_vppcalls.go#L1153-L1154
I'm curious about this link state. There are always if link is UP then interface is UP too, but I've tested with TAP, and it seems like enabling `linux-tap` does not mean that `vpp-tap` is enabled too. Where can I read about this links? If testing I need to know more about this, but right now for me link is only in case of TAP in vpp and other end in host.


Also about differences between `dumpInterfaces(ifIdxs ...uint32)` and `DumpInterfaceStates(ifIdxs ...uint32)`:
https://github.com/ligato/vpp-agent/blob/10336c576fae527496c070da4793fe86118d8b4f/plugins/vpp/ifplugin/vppcalls/vpp2001/dump_interface_vppcalls.go#L117-L118
and
https://github.com/ligato/vpp-agent/blob/10336c576fae527496c070da4793fe86118d8b4f/plugins/vpp/ifplugin/vppcalls/vpp2001/dump_interface_vppcalls.go#L394-L395 

Is this a bug too?

plugins/vpp/ifplugin/vppcalls/vpp2001/dump_interface_vppcalls.go
rewenset: Probably, because the "value 0" is not defined either, but it is returned in some cases. Actually, I've never seen `vpp_ifs.IF_STATUS_API_FLAG_ADMIN_UP` to be returned. For now it is always this three values
- 0
- IF_STATUS_API_FLAG_LINK_UP
- 3

milanlenco:
> Also about differences between dumpInterfaces(ifIdxs ...uint32) and DumpInterfaceStates(ifIdxs ...uint32)
> Is this a bug too?

yeah, the DumpInterfaceStates function seems to be wrong

> I'm curious about this link state

VPP and linux side of the same TAP should be really seen as two different interfaces (like a cable and interfaces are its ends). The admin status is like a logical switch that is set separately for each side. Link status is something that under normal circumstances should match between the sides. It basically tells if data can travel through the interfaces. If one of the sides is administratively disabled or if there is some issue with the connection, the link will be down on both sides.

rewenset:
```
  mp->flags |= (swif->flags & VNET_SW_INTERFACE_FLAG_ADMIN_UP) ?
    IF_STATUS_API_FLAG_ADMIN_UP : 0;
  mp->flags |= (hi->flags & VNET_HW_INTERFACE_FLAG_LINK_UP) ?
    IF_STATUS_API_FLAG_LINK_UP : 0;
  mp->flags = ntohl (mp->flags);
```
src: https://github.com/FDio/vpp/blob/master/src/vnet/interface_api.c#L211-L215

Now I understand why there is `3` returned. Is this still can be called a bug that's need to be in fdio jira?

plugins/vpp/ifplugin/vppcalls/vpp2001/dump_interface_vppcalls.go
```diff
@@ -1147,13 +1134,32 @@ func getVxLanGpeProtocol(p uint8) ifs.VxlanLink_Gpe_Protocol {
 	}
 }
 
-func ifAPIFlagState(flags vpp_ifs.IfStatusFlags) uint32 {
-	if flags == vpp_ifs.IF_STATUS_API_FLAG_ADMIN_UP {
-		return 1
-	} else if flags == vpp_ifs.IF_STATUS_API_FLAG_LINK_UP {
-		return 2
+func isAdminStateUp(flags vpp_ifs.IfStatusFlags) bool {
+	if flags&vpp_ifs.IF_STATUS_API_FLAG_ADMIN_UP == 0 {
```
golangcibot: S1008: should use 'return <expr>' instead of 'if <expr> { return <bool> }; return <bool>' (from `gosimple`)

plugins/vpp/ifplugin/vppcalls/vpp2001/dump_interface_vppcalls.go
```diff
@@ -1147,13 +1134,32 @@ func getVxLanGpeProtocol(p uint8) ifs.VxlanLink_Gpe_Protocol {
...
+		return false
+	}
+	return true
+}
+
+func isLinkStateUp(flags vpp_ifs.IfStatusFlags) bool {
+	if flags&vpp_ifs.IF_STATUS_API_FLAG_LINK_UP == 0 {
```
golangcibot: S1008: should use 'return <expr>' instead of 'if <expr> { return <bool> }; return <bool>' (from `gosimple`)

plugins/vpp/ifplugin/vppcalls/vpp2001/dump_interface_vppcalls.go
rewenset: oh my! Thanks ;)

plugins/vpp/ifplugin/vppcalls/vpp2001/dump_interface_vppcalls.go
rewenset: indeed

plugins/vpp/ifplugin/vppcalls/vpp2001/dump_interface_vppcalls.go
ondrej-fabry: Isn't it determined by the bits?

- `0` => admin down, link down
- `1` => admin up, link down
- `2` => admin down, link up
- `3` => admin up, link up

ondrej-fabry:
Here's the code that defines the interface flags, it indeed uses bits so interface can have multiple states at a time.

https://github.com/FDio/vpp/blob/33909777c63712ca397165cd92e7cc62208eb5c8/src/vnet/interface.h#L665-L690

plugins/vpp/ifplugin/vppcalls/vpp2001/dump_interface_vppcalls.go
rewenset: yes, bits is what matters. I've left a comment about it below

ondrej-fabry:
Here's code that might help to understand this:

```go
package main

import (
    "fmt"
)

const VNET_SW_INTERFACE_FLAG_ADMIN_UP = 1 << 0
const VNET_HW_INTERFACE_FLAG_LINK_UP = 1 << 0

// IfStatusFlags represents VPP binary API enum 'if_status_flags'.
type IfStatusFlags uint32

const (
    IF_STATUS_API_FLAG_ADMIN_UP IfStatusFlags = 1
    IF_STATUS_API_FLAG_LINK_UP  IfStatusFlags = 2
)

func main() {
    checkFlags(0, 0)
    checkFlags(1, 0)
    checkFlags(1, 1)
    checkFlags(0, 1) // this most likely does not make sense
    // output:
    // sw: 0 + hw: 0 = flags: 0 (bits: 00)
    // sw: 1 + hw: 0 = flags: 1 (bits: 01)
    // sw: 1 + hw: 1 = flags: 3 (bits: 11)
    // sw: 0 + hw: 1 = flags: 2 (bits: 10)
}

func checkFlags(swif, hwif uint32) {
    var  flags uint32
    flags |= adminUp(swif)
    flags |= linkUp(hwif)
    fmt.Printf("sw: %d + hw: %d = flags: %d (bits: %02b)\n", swif, hwif,flags,flags)
}

func adminUp(a uint32) uint32 {
    if (a & VNET_SW_INTERFACE_FLAG_ADMIN_UP)!=0 {
        return  uint32(IF_STATUS_API_FLAG_ADMIN_UP)
    }
    return 0
}


func linkUp(a uint32) uint32 {
    if (a &VNET_HW_INTERFACE_FLAG_LINK_UP)!=0 {
        return  uint32( IF_STATUS_API_FLAG_LINK_UP)
    }
    return 0
}


```


> https://play.golang.org/p/IvSuwytYnvG

milanlenco:
So yeah, not a VPP bug.
Were you able to find the VPP commit that changed the API to represent the link/admin status with bits? As part of that commit, did they change some other API in this way that we should be aware of?

PR #1481, opened 2019-09-27
test: Cleanup e2e tests

ondrej-fabry:
This PR cleans up code for E2E tests.

- using `SetDefaultEventuallyPollingInterval` and `SetDefaultEventuallyTimeout` instead of setting it for each `Eventually(..)` call
- using more consistent and opionately more readable assertions:
  - `Expect(err).To(BeNil())` -> `Expect(err).ToNot(HaveOccurred())`
  - `Expect(ping(..)).To(BeNil())` -> `Expect(ping(..)).To(Succeed())`
- compiling test agent as `vpp-agent.test` so it's ignored by git and setting version to `TEST_E2E` to make it clear it's special build
- using `set -x` to display the actual commands in `run_e2e.sh` script

### DIFF
--- a/tests/e2e/010_interfaces_test.go
+++ b/tests/e2e/010_interfaces_test.go
@@ -17,22 +17,16 @@ package e2e
 import (
 	"context"
 	"testing"
-	"time"
 
 	. "github.com/onsi/gomega"
 
-	"github.com/ligato/vpp-agent/api/models/linux/interfaces"
-	"github.com/ligato/vpp-agent/api/models/linux/namespace"
-	"github.com/ligato/vpp-agent/api/models/vpp/interfaces"
+	linux_interfaces "github.com/ligato/vpp-agent/api/models/linux/interfaces"
+	linux_namespace "github.com/ligato/vpp-agent/api/models/linux/namespace"
+	vpp_interfaces "github.com/ligato/vpp-agent/api/models/vpp/interfaces"
 	kvs "github.com/ligato/vpp-agent/plugins/kvscheduler/api"
 )
 
 // TODO: running downstream resync in-between restarts/re-creates seems to break stuff (for now commented out)
-
-const (
-	msUpdateTimeout  = time.Second * 6
-	recreateTimeout  = time.Second * 6
-)
 
 // connect VPP with a microservice via TAP interface
 func TestTapInterfaceConn(t *testing.T) {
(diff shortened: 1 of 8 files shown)


tests/e2e/run_e2e.sh
```diff
@@ -1,15 +1,24 @@
...
+go build -v -o ./tests/e2e/vpp-agent.test \
+  -ldflags "-X github.com/ligato/vpp-agent/vendor/github.com/ligato/cn-infra/agent.BuildVersion=TEST_E2E" \
+  ./cmd/vpp-agent
+
+# compile e2e test suite
+go test -c -o ./tests/e2e/e2e.test ./tests/e2e
+
+# start image
+# TODO: do not run docker image with pid=host,
+#  because any other vpp running now breaks tests
```
milanlenco: ACK, I'm going to try to get rid of --pid="host" and configure a common process namespace between the agent and microservices. Difficulty is that the PID of a microservice as given by docker daemon will not be the same as seen from the agent point of view, but it should be possible to find a workaround...

tests/e2e/e2e_test.go
```diff
@@ -285,10 +297,10 @@ func (ctx *testCtx) pingFromVPPClb(destAddress string) func() error {
 	}
 }
 
-func (ctx *testCtx) testConnection(fromMs, toMs, dstAddr, listenAddr string, port uint16, udp bool) error {
+/*func (ctx *testCtx) testConnection(fromMs, toMs, dstAddr, listenAddr string, port uint16, udp bool) error {
```
milanlenco: fyi: preparing NAT and ACL tests, this method will be provided with it.

tests/e2e/run_e2e.sh
ondrej-fabry: Perhaps using `--pid=container:$cid` should help here.

```
--pid=""  : Set the PID (Process) Namespace mode for the container,
             'container:<name|id>': joins another container's PID namespace
             'host': use the host's PID namespace inside the container
```
> https://docs.docker.com/engine/reference/run/#pid-settings---pid

tests/e2e/e2e_test.go
ondrej-fabry: Just wanted to avoid golangci bot from complaining that it's unused.

tests/e2e/run_e2e.sh
ondrej-fabry: Yeah, you will need to start microservices with `Pid: "container:<test-container-name>"`

tests/e2e/010_interfaces_test.go
```diff
@@ -246,39 +240,38 @@ func TestAfPacketInterfaceConn(t *testing.T) {
...
 
-	Expect(ctx.pingFromVPP(veth2IP)).ToNot(BeNil())
-	Expect(ctx.pingFromMs(msName, afPacketIP)).ToNot(BeNil())
+	Expect(ctx.pingFromVPP(veth2IP)).NotTo(Succeed())
+	Expect(ctx.pingFromMs(msName, afPacketIP)).NotTo(Succeed())
 
 	req = ctx.grpcClient.ChangeRequest()
 	err = req.Update(
 		afPacket,
 	).Send(context.Background())
```
rewenset: Maybe change this to one line too? As you did in the `tests/e2e/020_netalloc_test.go `. Same patterns occurs before and after this line

rewenset:
Maybe it would be better avoid package names like `utils` if it is possible? How about package named `http` with file `client.go` and `HTTPClient` -> `RestClient`.

ondrej-fabry:
> Maybe it would be better avoid package names like `utils` if it is possible? How about package named `http` with file `client.go` and `HTTPClient` -> `RestClient`.

Definitely agree here. This utils package will be removed after the update for agentctl which will have client API. Then the e2e tests will use that to manage agent.

tests/e2e/run_e2e.sh
milanlenco: fyi: this is the issue that we are facing with non-host pid namespaces: https://stackoverflow.com/questions/52021956/get-internal-docker-pid-by-system-pid

So for example, docker tells us that microservice "my-ms" has PID 5000, but inside the process namespace of the agent's container (that is shared with ms) the PID is something like 53. The mapping can be read from /proc, but mounting /proc from host is effectively the same as using --pid=host.

I'll dig into it more though, since avoiding --pid=host could give us more isolated environments...

ondrej-fabry: Oh, just realized that this can't be easily done with current implementation of microservice namespace reference. I opened an issue to continue this discussion: #1485

Issue #1503, opened 2019-10-07
Define duration in config files as string


ondrej-fabry:
This update in cn-infra: https://github.com/ligato/cn-infra/pull/408 makes it possible to define duration in config files as string and not only as integer in nanoseconds.

All config files should be updated to use this format so config files can be read and understood easily.

PR #1505, opened 2019-10-08
refactor: Update conf duration to string values

rewenset:
Issue #1503 

also updated cn-infra


tests/robot/resources/k8-yaml/vswitch.yaml
```diff
@@ -29,12 +29,12 @@ metadata:
 data:
   etcd.conf: |
     insecure-transport: true
-    dial-timeout: 1000000000
+    dial-timeout: 1
```
rewenset: damn. Will use force to fix it

PR #1522, opened 2019-10-15
feat: Add metrics registry

ondrej-fabry:


pkg/metrics/registry.go
```diff
@@ -14,15 +14,44 @@
...
-var registeredMetrics = make(map[string]func() proto.Message)
+	"github.com/ligato/vpp-agent/pkg/models"
+)
 
-func Register(pb proto.Message, fn func() proto.Message) {
-    models.
-    if _, ok := registeredMetrics; ok {
-        panic("duplicate registration for metrics %s", )
-    }
+var registeredMetrics = make(map[string]Retriever)
```
VladoLavor: This would need a mutex in case `Register` and `Retrieve` are called concurrently

Issue #1529, opened 2019-10-18
Support secure connections from AgentCTL


ondrej-fabry:
All communications used in AgentCTL should support secure connections.

Cconnections in AgentCTL
- [Etcd connection](https://github.com/ligato/vpp-agent/blob/dev/cmd/agentctl/client/client.go#L242) (to Etcd instance directly)
- [gRPC connection](https://github.com/ligato/vpp-agent/blob/dev/cmd/agentctl/client/client.go#L136) (to agent's gRPC server)
- [HTTP client](https://github.com/ligato/vpp-agent/blob/dev/cmd/agentctl/client/client.go#L150) (to agent's REST)

AgentCTL should accept/load single config file (like docker's `~/.docker/config.json`) where all these connections can setup keys/certs needed to establish secure connections.

Here are some references to docker code that can help to figure out how to setup secure connections on client side:
- [config](https://github.com/docker/cli/blob/master/cli/config/config.go)
- [flags](https://github.com/docker/cli/blob/master/cli/flags/common.go)
- [client](https://github.com/moby/moby/blob/master/client/client.go)

PR #1534, opened 2019-10-21
feat: TLS, HTTP basic auth and config file for AgentCTL

rewenset:
Issue #1529 
### Config File
Default Path: `~/.agentctl/config.yml`
Path to directory with config file may be changed with `--config-dir` flag

### Setup TLS
Example of conf file:
```sh
➜ cat ~/.agentctl/config.yml
kvdb-tls:
  cert-file: /home/user/certs/etcd1.pem
  key-file: /home/user/certs/etcd1-key.pem
  ca-file: /home/user/certs/ca.pem
  skip-verify: true

grpc-tls:
  cert-file: /home/user/certs/etcd1.pem
  key-file: /home/user/certs/etcd1-key.pem
  skip-verify: true

http-tls:
  disabled: true
➜ 
```

`insecure-tls: yes` in config file is the same as passing `--insecure-tls` flag. This configuration enables TLS with `skip-verify: true` (so client does not validate server's certificate) for all connections, even for disabled one (yes, it enables them but without using neither `cert-file`, nor `key-file`, and obviously nor `ca-file`). If TLS config was not disabled and `cert-file` with `key-file` are set, then `--insecure-tls` option will keep that as it is and use them for secure connections.

### Agent host configuration
0. Using defaults (I do that by set `config-dir` to some random path which not contains any configuration file for agentctl)
```sh
➜ agentctl --config-dir=/ status
HTTP request failed: Cannot connect to the agent at 127.0.0.1. Is the agent running?
```

1. Using config file only
```sh
➜ cat ~/.agentctl/config.yml
host: fromconfig
➜ agentctl status                    
HTTP request failed: error during connect: Get http://fromconfig:9191/readiness: dial tcp: lookup fromconfig: no such host
```

2. Using config file together with env variable.
```sh
➜ cat ~/.agentctl/config.yml
host: fromconfig
➜ AGENT_HOST=fromenv agentctl status
HTTP request failed: error during connect: Get http://fromenv:9191/readiness: dial tcp: lookup fromenv: no such host
```

3. Using config file together with env variable and with flag.
```sh
➜ cat ~/.agentctl/config.yml
host: fromconfig
➜ AGENT_HOST=fromenv agentctl --host fromflags status
HTTP request failed: error during connect: Get http://fromflags:9191/readiness: dial tcp: lookup fromflags: no such host
```

### Service label configuration
Config file line: `service-label: best-service-ever`
Env variable: `MICROSERVICE_LABEL= awesome-service`
Flag: `agentctl --service-label magic-service ...`

### gRPC and HTTP ports
Config file lines: 
```
grpc-port: 1010
http-port: 2020
```
Env variables: Not supported
Flag: `agentctl --grpc-port 1010 --http-port 2020 ...`

### ETCD endpoints
Config file lines: 
```
etcd-endpoints:
  - 1.1.1.1:1241
  - 2.2.2.2:3333
```
Env variable: `ETCD_ENDPOINTS= 1.1.1.1:1241,2.2.2.2:3333` (use `,` as separator)
Flag: `agentctl --debug --etcd-endpoints 1.1.1.1:1241 --etcd-endpoints 2.2.2.2:3333 ...`
or `agentctl --debug --etcd-endpoints 1.1.1.1:1241,2.2.2.2:3333 ...`

### Basic Auth for HTTP connection:
- Without credentials
```sh
➜ agentctl log ls agent                             
HTTP request failed: Error response from daemon: [401] Unauthorized.
➜ 
```
- Credentials provided with `--http-basic-auth` flag
```sh
➜ agentctl --http-basic-auth user:pass log ls agent      
LOGGER LEVEL 
agent  info  
➜ 
```
- Credentials provided with environment variable `AGENTCTL_HTTP_BASIC_AUTH`
```sh
➜ AGENTCTL_HTTP_BASIC_AUTH=user:pass agentctl log ls agent 
LOGGER LEVEL 
agent  info  
➜ 
```
- Credentials from config file:
```sh
➜  echo "http-basic-auth: user:pass" >> ~/.agentctl/config.yml
➜ agentctl log ls agent 
LOGGER LEVEL 
agent  info  
➜ 
```


rastislavszabo:
Maybe it would be easier to just reuse the exsiting [ConfigToClient](https://github.com/ligato/cn-infra/blob/master/db/keyval/etcd/config.go#L88) func from cn-infra?

Anyway, it would be useful if etcd endpoints could be passed via config file as well. Maybe even if any agentctl argument (e.g. service label) could be passed via config file?

cmd/agentctl/cli/config.go
```diff
@@ -0,0 +1,65 @@
...
+	"os"
+	"path/filepath"
+
+	"github.com/ligato/cn-infra/config"
+	"github.com/ligato/cn-infra/logging"
+)
+
+const (
+	configFileDir  = ".agentctl"
+	configFileName = "all.conf"
```
ondrej-fabry: Could the default config file name be rather named: `config.yml` to make it clear to user what the **file contains** (configuration) and which **data-serialization format** is used for its contents (YAML).

cmd/agentctl/cli/config.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+// ConfigFile represents info from ~/.agentctl/all.conf.
+type ConfigFile struct {
```
ondrej-fabry: I agree with comment from @rastislavszabo https://github.com/ligato/vpp-agent/pull/1534#issuecomment-544927983
> 
> 
> Anyway, it would be useful if etcd endpoints could be passed via config file as well. Maybe even if **any agentctl argument (e.g. service label) could be passed via config file**?

This config file definition (rather `Config`) should be summarizing all available configuration options that AgentCTL accepts/requires. There should be a `DefaultConfig()` function that would set the default values which should be used on line 56, then some of the options can get overridden by reading the config file contents, after that env vars and lastly the program arguments should be checked to [properly handle precedence order](https://github.com/spf13/viper#why-viper):

- default config
- config file (possibly system and then local)
- env vars
- program arguments

The Go library [viper](https://godoc.org/github.com/spf13/viper) might be really helpful here and provide simpler way to make this all work automatically.

rewenset: > a complete configuration for the AgentCTL

Sure, complete configuration sounds good. Will work on that too.

> The Go library viper might be really helpful

yeah, tab with Viper is opened in my browser from the beginning of work on this task, but for now I'm fighting for better TLS (I don't like that code from cn-infra). Definitely will look into Viper, but a little bit later, when I'll be working on a complete configuration and that order of different ways of configurating AgentCTL, I think.

cmd/agentctl/cli/config.go
rewenset: it was waiting for someone to suggest better name, thanks, I'll change it

rewenset:
@rastislavszabo thanks for checking out my work! Main problem with reusing is that the mentioned function not only creates connection to ETCD, but also tries to finish with configuration, and I don't want that. The second problem is with `tlsutils` from the etcd repo. I don't like neither `NewCert` nor `NewCertPool` even though it is three times copy-pasted:
- for kafka, I think: https://github.com/ligato/cn-infra/blob/0efd9f2edb1ebb87ab32a77adc0e96308bf85bf8/utils/clienttls/tlsutil.go#L35-L65
- for etcd: https://github.com/ligato/cn-infra/blob/0efd9f2edb1ebb87ab32a77adc0e96308bf85bf8/db/keyval/etcd/config.go#L127-L154
- for redis: https://github.com/ligato/cn-infra/blob/0efd9f2edb1ebb87ab32a77adc0e96308bf85bf8/db/keyval/redis/config.go#L39-L69

`NewCert` can be replaced with `LoadX509KeyPair` from `crypto/tls` and `NewCertPool` does part of the work already defined for `tls.Config` struct.
I'll try to propose more convenient way (Yet Another Way To Handle TLS :grin:)

cmd/agentctl/cli/flags.go
```diff
@@ -43,11 +42,9 @@ type ClientOptions struct {
 	PortHTTP     int
 	ServiceLabel string
 	Endpoints    []string
+	TLS          bool
 
-	// TODO: support TLS
-	TLS        bool
-	TLSVerify  bool
-	TLSOptions *tlsconfig.Options
```
ondrej-fabry: These options will be needed for providing security related options for connection (gRPC) to the agent and should not be used for the KVDB connection.

rewenset: So grpc is the only thing which can be configured via flags? Right now I'm using `--tls` flag to tell whether or not use TLS for different connections, for example if all files (cert, key and ca) are defined for kvdb in config file, but I want to connect to local etcd, so I don't need any TLS for now, but later I want to connect to remote etcd with auth enabled so I going to need those files. 
I'll rethink this with clear head tomorrow. Now I'm only about to push today's work and go home :relieved:

ondrej-fabry: Well, I guess we can keep this shared for now since the agentctl uses one connection per single run. However later when we start using viper any config option defined for configuration can be overriden automatically from [env vars](https://github.com/spf13/viper#working-with-environment-variables) and [flags](https://github.com/spf13/viper#working-with-flags) by using `BindEnv` and `BindFlags`.

cmd/agentctl/client/client.go
```diff
@@ -58,10 +65,12 @@ type Client struct {
 	addr     string
 	basePath string
 
-	grpcAddr     string
-	httpAddr     string
-	endpoints    []string
-	serviceLabel string
+	grpcAddr      string
+	grpcTLS       *TLSConfig
```
golangcibot: `grpcTLS` is unused (from `structcheck`)

cmd/agentctl/client/client.go
```diff
@@ -58,10 +60,13 @@ type Client struct {
...
 	basePath string
 
-	grpcAddr     string
-	httpAddr     string
-	endpoints    []string
-	serviceLabel string
+	grpcAddr      string
+	grpcTLS       *tls.Config
+	httpAddr      string
+	httpTLS       *tls.Config
```
golangcibot: `httpTLS` is unused (from `structcheck`)

cmd/agentctl/client/client.go
```diff
@@ -152,6 +149,14 @@ func (c *Client) GRPCConn() (*grpc.ClientConn, error) {
 
 // HTTPClient returns configured HTTP client.
 func (c *Client) HTTPClient() *http.Client {
+	if c.httpClient == nil {
+		tr := http.DefaultTransport.(*http.Transport).Clone()
```
rewenset: This `Clone()` method is new thing released in Go 1.13. Is it ok to use it?

cmd/agentctl/client/client.go
```diff
@@ -150,7 +151,24 @@ func (c *Client) GRPCConn() (*grpc.ClientConn, error) {
 // HTTPClient returns configured HTTP client.
 func (c *Client) HTTPClient() *http.Client {
 	if c.httpClient == nil {
-		tr := http.DefaultTransport.(*http.Transport).Clone()
+		// tr is copy-pasted definition of DefaultTransport from the net/http package
```
ondrej-fabry: You can actually use build tag for this. Check out this file from std package `sort` to see how.

https://github.com/golang/go/blob/0a338f75d4/src/sort/slice_go113.go

rewenset: woooow, didn't get it at first. So this is how you support different versions of Go (TIL). Good to know, I'll consider this

rewenset:
I'm done with TLS and it looks like [Viper](https://github.com/spf13/viper) requires big changes for Agenctl, so I think it will be better to continue (although I did not push anything) working on that in a new PR

cmd/agentctl/cli/flags.go
```diff
@@ -62,6 +57,12 @@ func (opts *ClientOptions) InstallFlags(flags *pflag.FlagSet) {
 	flags.IntVar(&opts.PortHTTP, "http-port", client.DefaultPortHTTP, "HTTP server port")
 	flags.IntVar(&opts.PortGRPC, "grpc-port", client.DefaultPortGRPC, "gRPC server port")
 	flags.StringSliceVarP(&opts.Endpoints, "etcd-endpoints", "e", etcdEndpoints, "Etcd endpoints to connect to, default from ETCD_ENDPOINTS env var")
+
+	flags.Bool("tls", false, "Use TLS for connections")
+	viper.BindPFlag("use-tls", flags.Lookup("tls"))
```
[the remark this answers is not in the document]
rewenset: There is address returned and I dont need it

cmd/agentctl/cli/flags.go
```diff
@@ -57,18 +43,29 @@ func NewClientOptions() *ClientOptions {
...
-	flags.IntVar(&opts.PortHTTP, "http-port", client.DefaultPortHTTP, "HTTP server port")
-	flags.IntVar(&opts.PortGRPC, "grpc-port", client.DefaultPortGRPC, "gRPC server port")
+	// TODO: consider using viper.AutomaticEnv with some prefix like `AGENTCTL`
+
+	flags.StringP("host", "H", client.DefaultAgentHost, "Address on which agent is reachable, default from AGENT_HOST env var")
+	viper.BindPFlag("host", flags.Lookup("host"))
+	viper.BindEnv("host", "AGENT_HOST")
+
+	flags.String("service-label", "", "Service label for specific agent instance, default from MICROSERVICE_LABEL env var")
+	viper.BindPFlag("service-label", flags.Lookup("service-label"))
```
golangcibot: Error return value of `viper.BindPFlag` is not checked (from `errcheck`)

cmd/agentctl/cli/flags.go
```diff
@@ -57,18 +43,29 @@ func NewClientOptions() *ClientOptions {
...
+	viper.BindEnv("service-label", "MICROSERVICE_LABEL")
```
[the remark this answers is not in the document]
rewenset: I know, it's ok

cmd/agentctl/cli/cli.go
```diff
@@ -187,14 +189,21 @@ func (cli *AgentCli) Initialize(opts *ClientOptions, ops ...InitializeOpt) error
...
 		client.WithHTTPPort(viper.GetInt("http-port")),
-		client.WithEtcdEndpoints(opts.Endpoints),
 	}
+
+	// Handle properly case when `etcd-endpoints` returned from environment variable
+	etcdEndp := viper.GetStringSlice("etcd-endpoints")
+	if len(etcdEndp) == 1 && strings.Contains(etcdEndp[0], ",") {
+		etcdEndp = strings.Split(etcdEndp[0], ",")
+	}
+	clientOpts = append(clientOpts, client.WithEtcdEndpoints(etcdEndp))
```
rewenset: Maybe don't allow multiple endpoints in environment variable and rename from `ETCD_ENDPOINTS` to `ETCD_ENDPOINT` when binding?

cmd/agentctl/cli/flags.go
```diff
@@ -57,18 +30,31 @@ func NewClientOptions() *ClientOptions {
...
+
+	flags.Int("http-port", client.DefaultPortHTTP, "HTTP server port")
+	viper.BindPFlag("http-port", flags.Lookup("http-port"))
+
+	flags.Int("grpc-port", client.DefaultPortGRPC, "gRPC server port")
+	viper.BindPFlag("grpc-port", flags.Lookup("grpc-port"))
+
+	flags.StringSliceP("etcd-endpoints", "e", []string{defaultEtcdEndpoint}, "Etcd endpoints to connect to, default from ETCD_ENDPOINTS env var")
+	viper.BindPFlag("etcd-endpoints", flags.Lookup("etcd-endpoints"))
+	viper.BindEnv("etcd-endpoints", "ETCD_ENDPOINTS")
```
golangcibot: Error return value of `viper.BindEnv` is not checked (from `errcheck`)

rastislavszabo:
One more ask, could you please add support for Basic HTTP Authentication for HTTP connections?
As used here: https://github.com/ligato/cn-infra/blob/master/rpc/rest/README.md#security
```
client-basic-auth:
  - "user:pass"
```

rewenset:
#### About E2E tests for secured HTTP connection:
Main problem is with that `waitUntilAgentReady` function, which is using bad (without TLS) httpClient
https://github.com/ligato/vpp-agent/blob/d6b37c3b7315271e145ad9c803f4e0d2e94962ac/tests/e2e/e2e_test.go#L481-L491
so we will never know when Agent is ready in case of configuring `HTTPS` for Agent. I'm not sure how to fix it in a correct way. I'm not sure if enabling TLS for HTTP and gRPC connection for all tests is a good idea, but if it is ok, let me know

### About E2E tests for secured ETCD connection:
We need ETCD running. I'm not sure how much time it would take, but I'll try to do that.

rewenset:
Before [`Speed up` commit](https://github.com/ligato/vpp-agent/pull/1534/commits/ad3246512e305eb7ddee02181684a1fda580f812):
```
e2e_test.go:509: agent ready, took 10.070035297s
```
After:
```
e2e_test.go:509: agent ready, took 102.859136ms
```
Super speed :zap:

cmd/agentctl/client/http.go
```diff
@@ -203,18 +203,20 @@ func (c *Client) checkResponseErr(serverResp serverResponse) error {
...
+		errorMsg = errorResponse.Message
 	} else {
-		errorMessage = strings.TrimSpace(string(body))
+		errorMsg = string(body)
 	}
 
-	return errors.Wrap(errors.New(errorMessage), "Error response from daemon")
+	errorMsg = fmt.Sprintf("[%d] %s", serverResp.statusCode, strings.TrimSpace(errorMsg))
+
+	return errors.Wrap(errors.New(errorMsg), "Error response from daemon")
```
rewenset: What is meant by the word "daemon" here? Is Agent a daemon?

tests/e2e/100_agentctl_test.go
```diff
@@ -303,3 +292,76 @@ func createFileWithContent(path, content string) error {
...
+		t.Logf("Setting `GRPC_CONFIG` back to %q", oldVal)
+		os.Setenv("GRPC_CONFIG", oldVal)
+	}(os.Getenv("GRPC_CONFIG"))
+	os.Setenv("GRPC_CONFIG", "/etc/grpc-secure.conf")
+
+	ctx := setupE2E(t)
+	defer ctx.teardownE2E()
+
+	t.Log("Try without any TLS")
+	stdout, stderr, err := ctx.execCmd(
```
golangcibot: ineffectual assignment to `stdout` (from `ineffassign`)

tests/e2e/100_agentctl_test.go
```diff
@@ -303,3 +292,76 @@ func createFileWithContent(path, content string) error {
...
+		"/agentctl", "--debug", "dump", "vpp.interfaces",
+	)
+	Expect(err).To(Not(BeNil()), "Should fail\n")
+	Expect(strings.Contains(stderr, "rpc error")).To(BeTrue(),
+		"Want in stderr: \n\"rpc error\"\nGot stderr: \n%s\n", stderr,
+	)
+	t.Log("PASSED")
+
+	t.Log("Try with TLS enabled via flag --tls, but without cert and key (note: server configured to check those files)")
+	stdout, stderr, err = ctx.execCmd(
```
golangcibot: ineffectual assignment to `stdout` (from `ineffassign`)

cmd/agentctl/client/http_client_go113.go
```diff
@@ -0,0 +1,34 @@
...
+// +build go1.13
+
+package client
+
+import (
+	"net/http"
+)
+
+// HTTPClient returns configured HTTP client.
+func (c *Client) HTTPClient() *http.Client {
```
ondrej-fabry: I think, the `HTTPClient` method should be defined once next to other `Client`'s methods. And these two ways of _cloning_ of transport for different go versions should be extracted into separate function `cloneHTTPTransport` and used like: 


```go
// HTTPClient returns configured HTTP client.
func (c *Client) HTTPClient() *http.Client {
	if c.httpClient == nil {
		tr := cloneHTTPTransport()
		tr.TLSClientConfig = c.httpTLS

		c.httpClient = &http.Client{
			Transport: tr,
		}
	}
	return c.httpClient
}
```

This way you minimize the redundant code while making it more clear what part is handled differently.

cmd/agentctl/client/tlsconfig/tlsconfig.go
```diff
@@ -0,0 +1,106 @@
...
+// 			if err != nil {
+// 				fmt.Printf("Error while creating TLS config: %v\n", err)
+// 				return
+// 			}
+// 			fmt.Println("TLS config is ready to use")
+//
+// 			// `tc` usage
+// 		}
+//
+
```
ondrej-fabry: > The convention is simple: to document a type, variable, constant, function, or even a package, write a regular comment directly preceding its declaration, with no intervening blank line...

Source: https://blog.golang.org/godoc-documenting-go-code

tests/e2e/100_agentctl_test.go
```diff
@@ -303,3 +292,76 @@ func createFileWithContent(path, content string) error {
 
 	return nil
 }
+
+func TestAgentCtlSecureGrpc(t *testing.T) {
+	// WARNING: Do not use grpc connection created in `setupE2E` in
+	// this test (though I don't know why you would but anyway).
+	// By default `grpc.Dail` is non-blocking and connecting happens
```
ondrej-fabry: Typo here.

cmd/agentctl/cli/config.go
```diff
@@ -0,0 +1,61 @@
...
+
+	p := filepath.Join(uhd, configFileDir)
+	logging.Debugf("default path to directory with agentctl's config is %q", p)
+	return p
+}
+
+// ReadConfig loads config using Viper.
+func ReadConfig() {
+	cfgFile := filepath.Join(viper.GetString("config-dir"), configFileName)
+	viper.SetConfigFile(cfgFile)
```
ondrej-fabry: I believe that setting _config file_ and _config type_ limits us to use only this specific file and type of config. However by using `AddConfigPath` and `SetConfigName` we could accept all of the [supported extensions](https://godoc.org/github.com/spf13/viper#pkg-variables) out of the box and even add multiple alternative paths.

cmd/agentctl/cli/cli.go
```diff
@@ -163,38 +166,100 @@ func (cli *AgentCli) Initialize(opts *ClientOptions, ops ...InitializeOpt) error
...
+		DefaultVersion: cli.client.Version(),
 	}
 	cli.initializeFromClient()
 	return nil
 }
 
-func newAPIClient(opts *ClientOptions) (client.APIClient, error) {
+func buildClientOptions() []client.Opt {
+	logging.Debug("----------------------------------------------------")
+	logging.Debug("Building client options")
```
ondrej-fabry: It would be easier and more explicit to use [`viper.Debug()`](https://godoc.org/github.com/spf13/viper#Debug) or [`viper.AllSettings()`](https://godoc.org/github.com/spf13/viper#AllSettings) for debugging actual config contents.

cmd/agentctl/cli/cli.go
ondrej-fabry: I would keep both though for backwards compatibility since we already use `ETCD_ENDPOINTS`.

cmd/agentctl/cli/cli.go
rewenset: Hpw about this?
```
	ReadConfig() // TODO: maybe move it elsewhere
	if opts.Debug {
		bConfig, err := json.MarshalIndent(viper.AllSettings(), "", " ")
		if err != nil {
			return fmt.Errorf("error while marshaling config to json: %v", err)
		}
		logging.Debugf("Current config: \n%s\n", string(bConfig))
	}
```

rastislavszabo:
> Opening some discussion... I think that the config structure should be defined as a `struct` and viper should only be used for merging all types of config input (config, flag, env..) and apply them to the `Config` struct using [`viper.Unmarshal`](https://godoc.org/github.com/spf13/viper#Viper.Unmarshal).

yeah, having the config in a `struct` would be better IMO as well

rewenset:
> Opening some discussion... I think that the config structure should be defined as a `struct` and viper should only be used for merging all types of config input (config, flag, env..) and apply them to the `Config` struct using [`viper.Unmarshal`](https://godoc.org/github.com/spf13/viper#Viper.Unmarshal).

In such case it would be easier to see what configuration can be applied and we may get rid of `viper.GetString...` and `viper.GetInt...`. Also for each field we can add a comment describing what flag and env variable it can be configured with, so this struct will be kind of documentation for configuring the AgentCtl. And there may be some methods for this Config struct such as whether or not use tls for kvdb and etc.

I agree and I'll start to work on that (shouldn't take much time). @ondrej-fabry may you approve this with :+1: on this comment?

rewenset:
It feels like endless PR :sparkles:
I hear tears dropping from face of everyone who advocates for merging into dev daily as CI practice  :grin:

ondrej-fabry:
> 
> 
> It feels like endless PR ✨
> I hear tears dropping from face of everyone who advocates for merging into dev daily as CI practice 😁

I feel you.. but better have it done properly at first than just increasing technical debt by putting "refactor.." task into backlog.

rewenset:
Configuration table:
```
| Name             | Flag             | Short Flag | Env                 | Conf           | Type      | Default            |
|------------------|------------------|------------|---------------------|----------------|-----------|--------------------|
| LigatoAPIVersion |                  |            | LIGATO_API_VERSION  |                | string    | "" (empty string)  |
| Host             | --host           | -H         | AGENT_HOST          | host           | string    | "127.0.0.1"        |
| ServiceLabel     | --service-label  |            | MICROSERVICE_LABEL  | service-label  | string    | "" (empty string)  |
| GRPCPort         | --grpc-port      |            |                     | grpc-port      | int       | 9111               |
| HTTPPort         | --http-port      |            |                     | http-port      | int       | 9191               |
| ETCDEndpoints    | --etcd-endpoints | -e         | ETCD_ENDPOINTS      | etcd-endpoints | []string  | ["127.0.0.1:2379"] |
| BasicAuth        | --basic-auth     |            | AGENTCTL_BASIC_AUTH | basic-auth     | string    | "" (empty string)  |
| UseTLS           | --tls            |            |                     | use-tls        | bool      | false              |
| GRPCSecure       |                  |            |                     | grpc-tls       | TLSConfig | nil                |
| HTTPSecure       |                  |            |                     | http-tls       | TLSConfig | nil                |
| KVDBSecure       |                  |            |                     | kvdb-tls       | TLSConfig | nil                |
|                                                                                                     |                    |
| Not in Config struct:                                                                               |                    |
|                  | --config-dir     |            |                     |                | string    | ""                 |
|                  | --debug          | -D         |                     |                | bool      |                    |
|                  | --version        | -v         |                     |                | bool      |                    |
|                  | --log-level      | -l         |                     |                | string    |                    |
|                  | --help           | -h         |                     |                | bool      |                    |
```

cmd/agentctl/cli/config.go
```diff
@@ -0,0 +1,107 @@
...
+	if err != nil {
+		return fmt.Sprintf("error while marshaling config to json: %v", err)
+	}
+
+	return string(bConfig)
+}
+
+// ShouldUseSecureGRPC returns whether or not to use TLS for GRPC connection.
+func (c *Config) ShouldUseSecureGRPC() bool {
+	return c.UseTLS == true && c.GRPCSecure != nil && !c.GRPCSecure.Disabled
```
golangcibot: S1002: should omit comparison to bool constant, can be simplified to `c.UseTLS` (from `gosimple`)

cmd/agentctl/cli/config.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+// ShouldUseSecureHTTP returns whether or not to use TLS for HTTP connection.
+func (c *Config) ShouldUseSecureHTTP() bool {
+	return c.UseTLS == true && c.HTTPSecure != nil && !c.HTTPSecure.Disabled
```
golangcibot: S1002: should omit comparison to bool constant, can be simplified to `c.UseTLS` (from `gosimple`)

cmd/agentctl/cli/config.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+// ShouldUseSecureKVDB returns whether or not to use TLS for KVDB connection.
+func (c *Config) ShouldUseSecureKVDB() bool {
+	return c.UseTLS == true && c.KVDBSecure != nil && !c.KVDBSecure.Disabled
```
golangcibot: S1002: should omit comparison to bool constant, can be simplified to `c.UseTLS` (from `gosimple`)

cmd/agentctl/cli/viper.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	viper.AddConfigPath(".")
+}
+
+// viperReadInConfig wraps viper.ReadInConfig with more logs.
+func viperReadInConfig() {
+	if err := viper.ReadInConfig(); err != nil {
+		if _, ok := err.(viper.ConfigFileNotFoundError); ok {
+			logging.Debug("unable to find config file")
```
ondrej-fabry: Better to print the actual error here too. It contains more specific context about where and what file was being accessed: https://github.com/spf13/viper/blob/master/viper.go#L115

ondrej-fabry:
That configuration table is slick! Would be nice to have a way to print this with something like `vpp-agent --print-config-table` 

..or perhaps even better with some **agentctl** command remotely?

rewenset:
> That configuration table is slick! Would be nice to have a way to print this with something like `vpp-agent --print-config-table`
> 
> ..or perhaps even better with some **agentctl** command remotely?

I'm glad you like it! I agree, it would be nice to have something like this printed by `agentctl`. But
- it was created manually and I'm not sure how to automate this 
- it was created to give high overview of current state of things

And also, I vote for separate command to do that, so with flags you may change columns printed and maybe level of verbosity. To existing columns, I would add "example" and "current value" and maybe "comment" too.

But it is another task and not suits in this PR! :)

Let's finish with this first, please. And, in a first place, by putting that "high overview" configuration table, I wanted to receive some feedback. Maybe we need rename something or add more ENV variables.

rastislavszabo:
Thanks for the table and all the work! I have some comments, but I would prefer finally merging this PR and addressing that (as well as the table you were discussing) in a separate PR.

- I would have one more config option candidate: ETCD dial-timeout (currently facing an issue with the hard-coded value)
- Is `BasicAuth` used only fot HTTP connections? If yes, what about naming it `HTTPBasicAuth`?
- is the `--tls ` flag really needed?

rewenset:
@rastislavszabo 
> ETCD dial-timeout

sure

> Is BasicAuth used only fot HTTP connections? If yes, what about naming it HTTPBasicAuth?

Yes, it is for HTTP only. Before now, I thought that basic auth is HTTP thing only. I'll rename it so it will be more clear

> is the --tls flag really needed?

I was thinking that user may not want to verify server's cert (or use host's CA set) so no need in CA file and server may not be configured to require client's cert so no need in cert and key files. In such case user just want that secure connection to work without any files and to achieve that you just add `--tls` flag and boom you're making **secure** requests now.

Somehow I've forgot about that and it's not working with latest changes. Thanks for pointing that out. In a current state there are two options: either remove that `--tls` completely or get that feature back.

<details><summary>Example of what I wanted to achieve:</summary>
<p>

- gRPC connection
```
➜ agentctl --config-dir=/ dump all  
rpc error: code = Unavailable desc = all SubConns are in TransientFailure, latest connection error: connection closed
➜ agentctl --config-dir=/ --tls dump all
KEY                             VALUE                               ORIGIN    METADATA                                                                           
config/vpp/v2/interfaces/tap1   [ligato.vpp.interfaces.Interface]   from-NB   map[IPAddresses:[10.10.10.10/24] SwIfIndex:2 TAPHostIfName:tap-3058328937 Vrf:0]   
                                name: "tap1"                                                                                                                     
                                type: TAP                                                                                                                        
                                enabled: true                                                                                                                    
                                ip_addresses: "10.10.10.10/24"                                                                                                   
                                tap: <                                                                                                                           
                                  version: 2                                                                                                                     
                                >                                                                                                                                
                                                                              
➜
```
- HTTP connection
```
➜ agentctl --config-dir=/ log ls agent       
HTTP request failed: Error response from daemon: [400] Client sent an HTTP request to an HTTPS server.
➜ agentctl --config-dir=/ --tls log ls agent
LOGGER LEVEL 
agent  info  
➜
```
- KVDB connection (indeed timeout will be nice thing to have)
```
➜ agentctl --config-dir=/ kvdb ls      
WARN[0003] Failed to connect to Etcd: context deadline exceeded  endpoints="[127.0.0.1:2379]" loc="etcd/bytes_broker_impl.go(60)" logger=etcd-client
connecting to KVDB failed: connecting to Etcd failed: context deadline exceeded
➜ agentctl --config-dir=/ --tls kvdb ls
/vnf-agent/vpp/foo
bar
➜
```
</p>
</details>

rastislavszabo:
> In such case user just want that secure connection to work without any files and to achieve that you just add `--tls` flag and boom you're making **secure** requests now.

Good idea, but maybe I would use different naming. E.g. with `curl`, you can achieve something similar with `--insecure` flag. What about `--insecure-tls` ?

rewenset:
> What about `--insecure-tls` ?

And by providing that flag, user will  know that no verification of server's certificate is done behind the scenes, so it is basically insecure security :grin:. Looks like a good naming choice :+1: 

And what if there is some TLS configuration? Will this flag override all connections to not verify server's certificate (and therefore skip CA file if set) and just send our cert (if it is set) for "authentication with HTTPS client certificates"?

rastislavszabo:
> And what if there is some TLS configuration? Will this flag override all connections to not verify server's certificate (and therefore skip CA file if set) and just send our cert (if it is set) for "authentication with HTTPS client certificates"?

yeah, this seems to be the best option

tests/e2e/resources/agentctl.conf
```diff
@@ -0,0 +1,16 @@
+use-tls: true
```
rastislavszabo: this option is already outdated, isn't it?

rewenset: it is, thanks

PR #1526, opened 2019-10-17
build: Migrate to go modules

loeliger:
Patch  set 2 -- addresses issues from first pull request.
- Also remove Gopkg.* files
- Fixed missing VPP in images
- Have 'make dep-check' run and check 'go mod tidy' for errors

PR #1541, opened 2019-10-24
Migrate to Go modules and vanity import path + organize proto files

ondrej-fabry:
Continue #1526 (rebased some conflicts)

This also involves little bit of extra setup before generating proto files.
All protobuf files moved under `proto/ligato/vpp-agent/` to follow [Buf setup](https://buf.build/docs/build-configuration) and to be:
- importable in other projects and still be under this repo
- make it easy to find for integration with other languages
- make imports in proto files more clear
- provide more context to proto descriptors to identify messages and services
- prevent conflicts in other projects (proto does not check files for duplicate registration!)
- make grpc reflection work properly out of the box


docker/dev/Dockerfile
```diff
@@ -3,27 +3,35 @@ FROM ${VPP_IMG} AS vppimg
 
 RUN dpkg-query -f '${Version}' -W vpp > /vpp/version
 
-FROM golang:1.12 AS verify-binapi
+FROM golang:1.13 AS verify-binapi
```
rewenset: awesome :star:

docker/dev/Dockerfile
```diff
@@ -86,7 +93,7 @@ COPY ./docker/dev/vpp.conf /etc/vpp/vpp.conf
 COPY ./docker/dev/init_hook.sh /usr/bin/
 
 # Build agent
-WORKDIR $GOPATH/src/github.com/ligato/vpp-agent
+WORKDIR /src/ligato/vpp-agent
 
 ARG VERSION
 ARG COMMIT
```
rewenset: Can you also add this lines, so only in case of a change in the dependencies, docker do `RUN go mod download` and use cached layers otherwise? It would help with rebuilding image with only source code changes.
```diff
 ARG COMMIT
 ARG DATE
 
+COPY go.mod go.sum ./
+RUN go mod download
+
 COPY . ./
```

ondrej-fabry:
It seems we will have to first change the import path to contain `/v2` suffix. Otherwise we are gonna break everyone that imports vpp-agent. See https://github.com/golang/go/wiki/Modules#releasing-modules-v2-or-higher for more info.

PR #1556, opened 2019-10-31
build: Cleanup proto generation and fix linter issues

ondrej-fabry:
- [x] fix some issues reported by protobuf linter
- [x] setup travis to run protobuf linter 
- [x] setup travis to run check for breaking changes
- [x] move protobuf related commands to `proto/buf.make`


pkg/models/item.go
```diff
@@ -71,14 +71,14 @@ func UnmarshalItem(item *api.Item) (proto.Message, error) {
 }
 
 // GetModelForItem returns model for given item.
-func GetModelForItem(item *api.Item) (RegisteredModel, error) {
+func GetModelForItem(item *api.Item) (KnownModel, error) {
```
VladoLavor: I think `RegisteredModel` fits more, but if you like it...

proto/ligato/generic/model.proto
```diff
@@ -1,51 +1,43 @@
...
-    ModelSpec spec = 1;
+    required ModelSpec spec = 1;
+
     // ProtoName is a name of protobuf message representing the model.
-    string proto_name = 2;
+    optional string proto_name = 2;
 
     message Option {
-        string key = 1;
+        required string key = 1;
```
VladoLavor: I see the meaning of the optional/required fields here, but wouldn't it be a better approach to solve it programatically? Using proto2 is fine, but switching just because of this is unnecessary.

scripts/checknodiffgenerated.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
```
VladoLavor: I know it's opinion-based, but we should use a convention in naming things - in this case, use underscores to separate words in the script name.

pkg/models/item.go
ondrej-fabry: Actually, in the situation when the models are retrieved from remote agent and not from local registry, it might be confusing to label them as registered.. 

To be honest, Im not certain either way..

proto/ligato/generic/model.proto
ondrej-fabry: This was not the reason behind using `proto2`. It was required to allow extension, but I moved it to a separate file to avoid defining this in `proto2`.

scripts/checknodiffgenerated.sh
ondrej-fabry: The name was taken from Buf project and since we have bunch of scripts without any spaces/underscores I left it like this for now. Agreed on the consistency though and it would be good idea to clear all the names sometime.

VladoLavor: I see, ok

PR #1559, opened 2019-10-31
fix: agentctl import with short keys

rastislavszabo:


ondrej-fabry:
This definitely needs some automated testing. I will merge for now to avoid blocking other tasks.

PR #1567, opened 2019-11-21
fix: Unnumbered interface name

mtirpak:
- During interface discovery, when any unnumbered interface referenced another interface without a tag, the VRF and IP address values could not be resolved and the delete operation failed.
This resolution was performed in the interface implementation but the name was only set later in the descriptor. This change moves the setting of the interface name earlier so that it is available already in the dump interfaces function.

- The above change uncovered a bug: 0 is a valid interface index (local0), but it was used as the multicast interface index for vxlan and gtpu tunnels when no multicast should be set. The integration tests were working only because the dump interfaces function returned an empty name for local0.

In my setup, gtpu_tunnel0 is an unnumbered interface using loop0. The unnumbered struct and and derived values were empty. The values were configured via vppctl and vpp-agent was unable to delete them:

```
      4. DELETE [DISCOVERED]:
          - key: config/vpp/v2/interfaces/UNTAGGED-loop0
          - value: { name:"UNTAGGED-loop0" type:SOFTWARE_LOOPBACK enabled:true phys_address:"de:ad:00:00:00:00" ip_addresses:"127.0.0.1/32"  } 
      5. DELETE [DERIVED DISCOVERED]:
          - key: vpp/interface/unnumbered/gtpu_tunnel0
          - value: {  } 
      6. DELETE [DERIVED DISCOVERED]:
          - key: vpp/interface/gtpu_tunnel0/vrf/from-interface/<invalid>
          - value: <EMPTY> 
      7. DELETE [DISCOVERED]:
          - key: config/vpp/v2/interfaces/gtpu_tunnel0
          - value: { name:"gtpu_tunnel0" type:GTPU_TUNNEL enabled:true unnumbered:<> gtpu:<src_addr:"172.16.16.2" dst_addr:"172.16.16.1" teid:100 decap_next:IP4 >  } 
...

ERRO[0000] failed to find interface gtpu_tunnel0         loc="descriptor/interface_vrf.go(173)" logger=vpp-ifplugin.interface-vrf-descriptor

```

### DIFF
--- a/plugins/vpp/ifplugin/descriptor/interface.go
+++ b/plugins/vpp/ifplugin/descriptor/interface.go
@@ -60,10 +60,6 @@ const (
 	//  - determined by much fits into the VPP interface tag (64 null-terminated character string)
 	logicalNameLengthLimit = 63
 
-	// prefix prepended to internal names of untagged interfaces to construct unique
-	// logical names
-	untaggedIfPreffix = "UNTAGGED-"
-
 	// suffix attached to logical names of dumped TAP interfaces with Linux side
 	// not found by Retrieve of Linux-ifplugin
 	tapMissingLinuxSideSuffix = "-MISSING_LINUX_SIDE"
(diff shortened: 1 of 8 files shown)


ondrej-fabry:
LGTM. Waiting for @milanlenco 's approval.

PR #1571, opened 2019-11-26
feat: vpp sr encap source address descriptor and configurator m…

artem-belov:
sr_set_source - global vpp configuration of sr encap source address

```
- key: config/vpp/srv6/v2/encapsourceaddress-global
            val: { encap_source:"2600:1f16:252:e900:542b:e8aa:8206:3d86"  }
```

Signed-off-by: Artem Belov <artem.belov@xored.com>


proto/ligato/vpp/srv6/srv6.proto
```diff
@@ -132,3 +132,8 @@ message Steering {
         string prefix_address = 2;       /* IPv4/IPv6 prefix address(CIRD format) of traffic destination. All traffic with given destination will be steered to given SR policy */
     }
 }
+
+/* IPv6 SR Set SRv6 encapsulation source */
+message EncapSourceAddress {
```
ondrej-fabry: I think that `EncapSourceAddress` message should rather be named in a way that it can be extended in the future when possibly some other _srv6 related_ global settings is needed. Perhaps `SRv6Global`?  What do you think? @artem-belov

artem-belov: I agree

PR #1581, opened 2019-12-06
feat: Agentctl output format

ondrej-fabry:

### DIFF
--- a/cmd/agentctl/commands/model.go
+++ b/cmd/agentctl/commands/model.go
@@ -17,7 +17,6 @@ package commands
 import (
 	"bytes"
 	"context"
-	"encoding/json"
 	"fmt"
 	"io"
 	"path"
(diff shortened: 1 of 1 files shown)


cmd/agentctl/commands/model.go
```diff
@@ -168,8 +167,7 @@ func newModelInspectCommand(cli agentcli.Cli) *cobra.Command {
 			return runModelInspect(cli, opts)
 		},
 	}
-	// TODO: add support for custom formatting instead of json
-	//cmd.Flags().StringVar(&opts.Format, "format", "", "Format for the output")
+	cmd.Flags().StringVarP(&opts.Format, "format", "f", "", "Format for the output")
```
rewenset: why not use "json" as default value here (instead of empty string)?

ondrej-fabry: To make the default output type configurable globally in the future for all commands. Config for agentctl might define default as `yaml` then empty string will default to YAML format instead of JSON.

PR #1595, opened 2019-12-19
refactor: NAT Interface and AddressPool API changes

rastislavszabo:
This PR moves NAT interfaces and NAT address pools from the global NAT44 config into separate NB API entities.

At the same time, it supports full backward compatibility with the old (deprecated) API. This is done by keeping both old (global - deprecated) and new NAT address pool & interface descriptors, which share the same VPPcalls. Once the deprecated API is to be definitively removed, the old descriptors can be just deleted.


clientv2/vpp/dbadapter/data_change_db.go
```diff
@@ -167,6 +167,18 @@ func (dsl *PutDSL) DNAT44(nat44 *nat.DNat44) vppclient.PutDSL {
 	return dsl
 }
 
+// NAT44Interface adds a request to create or update NAT44 interface configuration.
+func (dsl *PutDSL) NAT44Interface(natIf *nat.Nat44Interface) vppclient.PutDSL {
+	dsl.parent.txn.Put(models.Key(&nat.Nat44Interface{}), natIf)
```
ondrej-fabry: Shouldn't this be `models.Key(natIf)`?

clientv2/vpp/dbadapter/data_change_db.go
```diff
@@ -167,6 +167,18 @@ func (dsl *PutDSL) DNAT44(nat44 *nat.DNat44) vppclient.PutDSL {
...
+	return dsl
+}
+
+// NAT44AddressPool adds a request to create or update NAT44 address pool.
+func (dsl *PutDSL) NAT44AddressPool(pool *nat.Nat44AddressPool) vppclient.PutDSL {
+	dsl.parent.txn.Put(models.Key(&nat.Nat44AddressPool{}), pool)
```
ondrej-fabry: Same case  here.

proto/ligato/vpp/nat/nat.proto
```diff
@@ -75,7 +75,26 @@ message DNat44 {
     repeated IdentityMapping id_mappings = 3;
 }
 
-message VirtualReassembly {                 /* NAT virtual reassembly */
+/* Local network interfaces enabled for NAT44. */
+message Nat44Interface {
+    string name = 1;                        /* (logical) Interface name. */
```
ondrej-fabry: Comments in proto should be line comments located above the particular field/message. Otherwise they don't end up in generated file.

clientv2/vpp/dbadapter/data_change_db.go
rastislavszabo: it should, thx!

proto/ligato/vpp/nat/nat.proto
rastislavszabo: Good idea. I was keeping the style of comments in the existing proto. I can refactor it completely.

rewenset:
Currently `vpp-agent/examples/kvscheduler/nat` shows:
```
WARN[0002] NatInterfaces are deprecated in global NAT44 config, use separate Nat44Interface entries.  loc="descriptor/nat44_global.go(118)" logger=vpp-natplugin.nat44-global-descriptor
WARN[0002] AddressPool is deprecated in global NAT44 config, use separate Nat44AddressPool entries.  loc="descriptor/nat44_global.go(121)" logger=vpp-natplugin.nat44-global-descriptor
```
Probably, examples should be updated too if possible.
Same applies to `vpp-agent/examples/localclient_vpp/nat`

rastislavszabo:
> For me, `TestSourceNAT` fails and `TestNATStaticMappings` just prints warnings.

Thanks @rewenset , tests should not fail even if using old API, since this aims to be backward compatible. It seems that I forgot to setup [this dependency in address descriptor](https://github.com/ligato/vpp-agent/blob/5903611b1e71ae005067bc00eb39ffe3adfe07e8/plugins/vpp/natplugin/descriptor/nat44_address_pool.go#L70-L71)

Will update e2e-tests as well.

PR #1602, opened 2020-01-17
feat: Add L3XC support

ondrej-fabry:


plugins/vpp/l3plugin/descriptor/route.go
```diff
@@ -21,10 +21,10 @@ import (
 	"strings"
 
 	"github.com/golang/protobuf/proto"
-	"github.com/pkg/errors"
-
 	"github.com/ligato/cn-infra/logging"
 	"github.com/ligato/cn-infra/utils/addrs"
+	"github.com/pkg/errors"
```
rewenset: but why did you move it to group with cn-infra packages?

ondrej-fabry: I'm using `goimports` and I've got Go imports setup in Goland in 3 groups: std, other modules, current module.

rewenset: oh, thanks, sorry for bothering :)

ondrej-fabry: It's no bother ;)

proto/ligato/vpp/l3/l3xc.proto
```diff
@@ -0,0 +1,23 @@
+syntax = "proto3";
+
+package ligato.vpp.l3;
+
+option go_package = "go.ligato.io/vpp-agent/v3/proto/ligato/vpp/l3;vpp_l3";
+
+message L3XC {
```
rastislavszabo: I would suggest to use the name `L3XConnect` - so that:
- users of the API have some more clue on what it actually is (especially if it lacks any doc comments)
- it is more in line with [L2 xconnect](https://github.com/ligato/vpp-agent/blob/master/proto/ligato/vpp/l2/xconnect.proto) naming

proto/ligato/vpp/l3/l3xc.proto
```diff
@@ -0,0 +1,23 @@
...
+    string interface = 1;
+
+    enum Protocol {
+        IPV4 = 0;
+        IPV6 = 1;
+    }
+    Protocol protocol = 2;
+
+    message Path {
+        string interface = 1;
```
rastislavszabo: Maybe we should consider keeping the naming in line with [Route API](https://github.com/ligato/vpp-agent/blob/master/proto/ligato/vpp/l3/route.proto#L24-L25):
- `outgoing_interface`
- `next_hop_addr`

PR #1606, opened 2020-01-20
fix: Handling disabled VPP plugins

ondrej-fabry:


plugins/vpp/ifplugin/ifplugin.go
```diff
@@ -52,6 +54,11 @@ import (
 	_ "go.ligato.io/vpp-agent/v3/plugins/vpp/ifplugin/vppcalls/vpp2001_324"
 )
 
+func init() {
+	// this adds ErrPluginDisabled to non-retryable errors
+	kvscheduler.NonRetryableErrors = append(kvscheduler.NonRetryableErrors, vppclient.ErrPluginDisabled)
```
rewenset: why this appending is here and not somewhere near to declaration of `ErrPluginDisabled` ?

ondrej-fabry: Because it would make `plugins/vpp` package depend on `kvscheduler`.

plugins/vpp/vpp.go
```diff
@@ -56,17 +56,32 @@ func FindCompatibleBinapi(ch CompatibilityChecker) (binapi.Version, error) {
...
 		if err := ch.CheckCompatiblity(msgs...); err == nil {
 			logging.Debugf("found compatible binapi version: %v", version)
 			return version, nil
 		} else if ierr, ok := err.(*govppapi.CompatibilityError); ok {
-			logging.Debugf("binapi version %-15v incompatible: %d/%d incompatible messages", version, len(ierr.IncompatibleMessages), len(msgs))
+			logging.Debugf("binapi version %-15v incompatible: %d/%d incompatible messages",
+				version, len(ierr.IncompatibleMessages), len(msgs))
+
+			if mostCompatible.version == "" || mostCompatible.incompatible > len(ierr.IncompatibleMessages) {
+				mostCompatible.version = version
```
rewenset: finally I've found something. `mostCompatible` can be version without compatible messages. Please tell me that I'm right and it is something :)

ondrej-fabry: Hm.. actually I think you're on point! I should check if at least some messages are compatible.

plugins/vpp/vpp.go
```diff
@@ -64,19 +66,25 @@ func FindCompatibleBinapi(ch CompatibilityChecker) (binapi.Version, error) {
...
+		if err := ch.CheckCompatiblity(msgList.Plugins.AllMessages()...); err == nil {
 			logging.Debugf("found compatible binapi version: %v", version)
 			return version, nil
 		} else if ierr, ok := err.(*govppapi.CompatibilityError); ok {
 			logging.Debugf("binapi version %-15v incompatible: %d/%d incompatible messages",
 				version, len(ierr.IncompatibleMessages), len(msgs))
-
-			if mostCompatible.version == "" || mostCompatible.incompatible > len(ierr.IncompatibleMessages) {
+			incompatible := len(ierr.IncompatibleMessages)
+			if (mostCompatible.version == "" || mostCompatible.incompatible > incompatible) && incompatible < len(msgs) {
```
rewenset: `incompatible` will be always less than length of all messages, because here we check for compatibility only `msgList.Plugins.AllMessages()`

plugins/vpp/vpp.go
```diff
@@ -64,19 +66,25 @@ func FindCompatibleBinapi(ch CompatibilityChecker) (binapi.Version, error) {
 	}{}
 	for version, msgList := range binapi.Versions {
 		msgs := msgList.AllMessages()
```
rewenset: no need to define `msgs` right in the beginning

plugins/vpp/vpp.go
```diff
@@ -64,19 +66,25 @@ func FindCompatibleBinapi(ch CompatibilityChecker) (binapi.Version, error) {
...
-		if err := ch.CheckCompatiblity(msgs...); err == nil {
+		if err := ch.CheckCompatiblity(msgList.Core.AllMessages()...); err == nil {
```
rewenset: maybe check `err != nil` and continue. In this case no need of `else` block

plugins/vpp/vpp.go
```diff
@@ -64,19 +66,25 @@ func FindCompatibleBinapi(ch CompatibilityChecker) (binapi.Version, error) {
...
+		} else {
+			logging.Debugf("binapi version %v core check failed: %v", version, err)
+			continue
+		}
+		if err := ch.CheckCompatiblity(msgList.Plugins.AllMessages()...); err == nil {
 			logging.Debugf("found compatible binapi version: %v", version)
 			return version, nil
 		} else if ierr, ok := err.(*govppapi.CompatibilityError); ok {
 			logging.Debugf("binapi version %-15v incompatible: %d/%d incompatible messages",
 				version, len(ierr.IncompatibleMessages), len(msgs))
```
rewenset: `incompatible` can be defined before log message and used here instead of `len(ierr.IncompatibleMessages)`

plugins/vpp/handlers.go
```diff
@@ -79,6 +79,12 @@ func (h *Handler) GetCompatibleVersion(c Client) (*HandlerVersion, error) {
...
 	// try compatible binapi version first
+	if ver := c.PreferredVersion(); ver != "" {
+		if v, ok := h.versions[ver]; ok {
+			logging.Debugf("VPP handler %s COMPATIBLE with preferred version: %s", h.desc.Name, v.Version)
+			return v, nil
+		}
+	}
 	if ver, _ := FindCompatibleBinapi(c); ver != "" {
 		if v, ok := h.versions[ver]; ok {
 			logging.Debugf("VPP handler %s COMPATIBLE with version: %s", h.desc.Name, v.Version)
```
rewenset: maybe rewrite this message, but not I'm not sure

PR #1608, opened 2020-01-22
fix: shellcheck almost happy

rewenset:
:warning:There were 4 problems with adding double quotes to substitution (i.e. `${SMTH-}` -> `"${SMTH-}"`). Not sure about everything else. :warning:

Maybe this PR can be something to start with or not. Maybe do it manually is a better choice. Anyway 

This PR **reduces number of shell script issues from 345 to 105 using auto fix feature of [shellcheck](https://github.com/koalaman/shellcheck)**.

### The command itself:
```
docker run --rm -v "$PWD:/mnt" koalaman/shellcheck-alpine:latest find /mnt -name "*.sh" -exec shellcheck -f diff {} '+' | sed 's/\/mnt/\./g' | git apply
```
### And explanation:
Multiline to better look on it:
```
docker run --rm -v "$PWD:/mnt" koalaman/shellcheck-alpine:latest 
        find /mnt -name "*.sh" 
                -exec shellcheck -f diff {} '+' 

| sed 's/\/mnt/\./g' 
| git apply
```
first start container from "**koalaman/shellcheck-alpine:latest**" and mount there whole project. Command for this container will be "find" which will find (yeah, it will) all "*.sh" files in project's ("/mnt" in this case) directory. Also, "find" command can do "exec" and that's the point where "shellcheck" jumps in. Latest version of "shellcheck" can do **[autofix](https://github.com/koalaman/shellcheck/issues/1220)** with output in diff mode (-f diff). And "+" is for "find" and it means that just pass all file paths to command and call it once instead of calling command ("shellcheck" in this case) for each file (or something like that).
After all of that is done, the output is diff, but file paths are wrong. So "sed" will fix it. And now it is ready for "git apply".

<details><summary>Maybe useful output from shellcheck</summary>
<p>

```sh
➜ docker run --rm koalaman/shellcheck-alpine:latest shellcheck -V                                                
ShellCheck - shell script analysis tool
version: 0.7.0
license: GNU General Public License, version 3
website: https://www.shellcheck.net

➜ docker run --rm koalaman/shellcheck-alpine:latest shellcheck --help
Usage: shellcheck [OPTIONS...] FILES...
  -a                  --check-sourced            Include warnings from sourced files
  -C[WHEN]            --color[=WHEN]             Use color (auto, always, never)
  -i CODE1,CODE2..    --include=CODE1,CODE2..    Consider only given types of warnings
  -e CODE1,CODE2..    --exclude=CODE1,CODE2..    Exclude types of warnings
  -f FORMAT           --format=FORMAT            Output format (checkstyle, diff, gcc, json, json1, quiet, tty)
                      --list-optional            List checks disabled by default
                      --norc                     Don't look for .shellcheckrc files
  -o check1,check2..  --enable=check1,check2..   List of optional checks to enable (or 'all')
  -P SOURCEPATHS      --source-path=SOURCEPATHS  Specify path when looking for sourced files ("SCRIPTDIR" for script's dir)
  -s SHELLNAME        --shell=SHELLNAME          Specify dialect (sh, bash, dash, ksh)
  -S SEVERITY         --severity=SEVERITY        Minimum severity of errors to consider (error, warning, info, style)
  -V                  --version                  Print version information
  -W NUM              --wiki-link-count=NUM      The number of wiki links to show, when applicable
  -x                  --external-sources         Allow 'source' outside of FILES
                      --help                     Show this usage summary and exit

➜
```

</p>
</details>

### DIFF
--- a/tests/integration/vpp_integration.sh
+++ b/tests/integration/vpp_integration.sh
@@ -25,7 +25,7 @@ echo -e " VPP Integration Test - \e[1;33m${vppver}\e[0m"
 echo "============================================================="
 
 # run integration test
-if docker exec -i "$cid" /vpp-integration.test "$*"; then
+if docker exec -i "$cid" /vpp-integration.test "$@"; then
 	echo >&2 "-------------------------------------------------------------"
 	echo >&2 -e " \e[32mPASSED\e[0m (took: ${SECONDS}s)"
 	echo >&2 "-------------------------------------------------------------"
--- a/tests/integration/vpp_integration.sh
+++ b/tests/integration/vpp_integration.sh
@@ -8,7 +8,7 @@ go test -c ./tests/integration/vpp -o ./tests/integration/vpp/vpp-integration.te
 cid=$(docker run -d -it \
 	-v $(pwd)/tests/integration/vpp/vpp-integration.test:/vpp-integration.test:ro \
 	--label vpp.integration.test="$*" \
-	"${DOCKER_ARGS-}" \
+	${DOCKER_ARGS-} \
 	"$VPP_IMG" bash)
 
 on_exit() {
--- a/tests/integration/vpp_integration.sh
+++ b/tests/integration/vpp_integration.sh
@@ -8,7 +8,7 @@ go test -c ./tests/integration/vpp -o ./tests/integration/vpp/vpp-integration.te
 cid=$(docker run -d -it \
 	-v $(pwd)/tests/integration/vpp/vpp-integration.test:/vpp-integration.test:ro \
 	--label vpp.integration.test="$*" \
-	${DOCKER_ARGS-} \
+	"${DOCKER_ARGS-}" \
 	"$VPP_IMG" bash)
 
 on_exit() {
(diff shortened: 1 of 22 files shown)


tests/integration/vpp_integration.sh
```diff
@@ -25,7 +25,7 @@ echo -e " VPP Integration Test - \e[1;33m${vppver}\e[0m"
 echo "============================================================="
 
 # run integration test
-if docker exec -i "$cid" /vpp-integration.test $*; then
+if docker exec -i "$cid" /vpp-integration.test "$*"; then
```
ondrej-fabry: Will this one really work with `$*` in quotes?

ondrej-fabry:
Could you merge `ci-workflow` branch here? So we can see shellcheck output.

tests/integration/vpp_integration.sh
rewenset: yeah, without those quotes it works, but with them I get:
```sh
➜ VPP_IMG=ligato/vpp-base:19.08 ./tests/integration/vpp_integration.sh -test.run TestPing -test.v
=============================================================
 VPP Integration Test - 19.08.1-353~g418b22593~b80
=============================================================
flag provided but not defined: -test.run TestPing -test.v
Usage of /vpp-integration.test:
  -debug
        Turn on debug mode.
  -retry uint
        Number of VPP connect retries (default 3)
[...]
```

but it also works with `"$@"`, so actually I'm not sure why their autofix changes `$*` to `"$*"` and not to `"$@"`.

PR #1610, opened 2020-01-28
feat: IP flow hash settings support

rastislavszabo:


plugins/vpp/l3plugin/descriptor/vrf_table.go
```diff
@@ -167,3 +203,14 @@ func getVrfTableLabel(vrfTable *l3.VrfTable) string {
 	}
 	return vrfTable.Label
 }
+
+// defaultVrfFlowHashSettings returns default flow hash settings (implicitly existing if not set).
+func defaultVrfFlowHashSettings() *l3.VrfTable_FlowHashSettings {
+	return &l3.VrfTable_FlowHashSettings{
+		UseSrcIp:    true,
```
ondrej-fabry: Could you also list other fields that are `false` by default? Just for clarity for readers and possibility of new fields in future.

PR #1614, opened 2020-01-31
feat: Switch to proto-gen-go v1.3.3 to start using grpc.ClientC…

edwarnicke:
proto-gen-go v1.3.3 starts to use grpc.ClientConnInterface instead
of the struct grpc.ClientConn .  This makes it possible to do
Mocking for unit testing and is super super super cool :)


ondrej-fabry:
@edwarnicke thanks!

PR #1622, opened 2020-02-12
improvement: Expose GoVPP stats in govppmux API

rastislavszabo:

### DIFF
--- a/plugins/govppmux/plugin_api_govppmux.go
+++ b/plugins/govppmux/plugin_api_govppmux.go
@@ -15,14 +15,19 @@
 package govppmux
 
 import (
+	"go.ligato.io/vpp-agent/v3/pkg/metrics"
 	"go.ligato.io/vpp-agent/v3/plugins/govppmux/vppcalls"
 	"go.ligato.io/vpp-agent/v3/plugins/vpp"
+	"go.ligato.io/vpp-agent/v3/proto/ligato/govppmux"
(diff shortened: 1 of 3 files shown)


ondrej-fabry:
@rastislavszabo I'm not sure putting statistics/metrics as part of govppmux plugin (or any other plugin) API is a proper solution, especially because the stats are global and not coupled with the plugin. 

Could you help clarify the reasoning behind this change? There already was exported function `GetStats()` for retrieving the stats data.

rastislavszabo:
@ondrej-fabry The idea was to expose govppp binary api counters/stats to make them consumable programmatically, not only via REST. These are coupled with the govppmux plugin, Would be exposing everything except `govppmux.Metrics` more acceptable to you? E.g.:
```go
type BinAPIStats struct {
	Errors      metrics.Calls
	AllMessages metrics.CallStats
	Messages    metrics.Calls
	Replies     metrics.Calls
}
```

Or is it not acceptable at all?

ondrej-fabry:
I was not referring to retrieving stats via REST. You can access the govppmux stats programmatically with the current implementation already, using the exported function `GetStats()`. This PR only adds a new method for Plugin that calls `GetStats()` and returns just it, which in essence does not make much sense since it's global data.

rastislavszabo:
right, OK it's not really needed to have in included in API

PR #1626, opened 2020-02-21
feat: Add promethues metrics to govppmux

ondrej-fabry:
This PR contains:
- add prometheus metrics to govppmux and telemetry (more to come)
- fix parsing VPP version (strip zero bytes)
- added integration test case for parsing VPP version
- fix race condition when for concurrent stats retrievval from govppmux plugin
- update version package and some minor code cleanups


pkg/version/version.go
```diff
@@ -64,22 +68,24 @@ func Short() string {
 	return fmt.Sprintf(`%s %s`, app, version)
 }
 
-func BuiltStamp() string {
-	return fmt.Sprintf("%s (%s)", buildTime.Format(time.UnixDate), timeAgo(buildTime))
+func BuiltOn() string {
+	stamp := fmt.Sprintf("%s", buildTime.Format(time.UnixDate))
```
golangcibot: S1025: the argument is already a string, there's no need to use fmt.Sprintf (from `gosimple`)

PR #1635, opened 2020-03-03
test: Fix occasional e2e failure

ondrej-fabry:


tests/e2e/010_interfaces_test.go
```diff
@@ -336,7 +336,7 @@ func TestAfPacketWithLogicalReference(t *testing.T) {
 	).Send(context.Background())
 	Expect(err).ToNot(HaveOccurred())
 
-	Expect(ctx.getValueState(afPacket)).To(Equal(kvscheduler.ValueState_CONFIGURED))
+	Eventually(ctx.getValueStateClb(afPacket)).Should(Equal(kvscheduler.ValueState_CONFIGURED))
```
milanlenco: Hmm, it is strange this was causing failures. With reference to logical name of VETH1, the AF-PACKET should be ready for configuration already in that first transaction (with host name reference, AF-PACKET is created only after notification from netlink is received few transactions later). Do you still have a link where this test failed?

ondrej-fabry: Here are some recent failures one travis:
- https://travis-ci.com/ligato/vpp-agent/jobs/292763530
  `--- FAIL: TestAfPacketWithLogicalReference (5.09s)`
- https://travis-ci.com/ligato/vpp-agent/jobs/291474348
  `--- FAIL: TestTapInterfaceConn (70.42s)`
  `--- FAIL: TestAfPacketWithLogicalReference (5.32s)`

What might be happening is that the detection of microservice (SB notification) comes after the request to update, since there is nothing in this test (maybe others too) that would wait for the microservice to be discovered by agent.

milanlenco: Right, it is the pending microservice indeed. Thanks

PR #1644, opened 2020-04-02
perf: Performance enhancement for adding many rules to Linux IPtables chain

fgschwan:
The performance enhancement is based on usage of `iptables-save`/`iptables-restore`. It can speed up insertion of rules into chain significantly, but it is different approach then use of more generic `iptables` tool.

Due to changing `iptables-save`-exported data that are in not well documented format (not an API in classic sense), i made this performance enhancement optional and by default turned off.

The usage of this method is regulated by rule count that we want to insert into chain. So inserting few rules that can be quick also with generic `iptables` tool can benefit from stability of `iptables` tool API. Inserting many rules will switch to performance method and can be done much more quickly, but with possibility of future incompatibility break. The data export format can change without notice, but i don't think that it will break anytime soon, because:
a) this format seems to quite mature and stable (didn't find signs of changes in last years)
b) i don't do massive data parsing - i just insert data into one place

### DIFF
--- a/plugins/linux/iptablesplugin/descriptor/rulechain.go
+++ b/plugins/linux/iptablesplugin/descriptor/rulechain.go
@@ -15,12 +15,13 @@
 package descriptor
 
 import (
+	"bytes"
+	"fmt"
 	"strings"
 
 	"github.com/golang/protobuf/proto"
 	"github.com/pkg/errors"
 	"go.ligato.io/cn-infra/v2/logging"
-
 	kvs "go.ligato.io/vpp-agent/v3/plugins/kvscheduler/api"
 	ifdescriptor "go.ligato.io/vpp-agent/v3/plugins/linux/ifplugin/descriptor"
 	"go.ligato.io/vpp-agent/v3/plugins/linux/iptablesplugin/descriptor/adapter"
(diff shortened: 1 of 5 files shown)


plugins/linux/iptablesplugin/descriptor/rulechain.go
```diff
@@ -208,11 +212,43 @@ func (d *RuleChainDescriptor) Create(key string, rch *linux_iptables.RuleChain)
 	}
 
 	// append all rules
-	for _, rule := range rch.Rules {
-		err := d.ipTablesHandler.AppendRule(protocolType(rch), tableNameStr(rch), chainNameStr(rch), rule)
-		if err != nil {
-			d.log.Errorf("Error by appending iptables rule: %v", err)
-			break
+	if len(rch.Rules) > 0 {
```
ondrej-fabry, line 215: Does this implementation of optimized rule appending really need to be in descriptor? From descriptor's perspective it should not really matter _how_ the rules are added. 

The handler API should hide implementation details like this behind _abstracted_ interface, so it can be swapped out with different handler implementation in the future without affecting higher-level code.

**EXAMPLE:** 
Someone adds new handler implementation for `IPTablesAPI` which instead of invoking `iptables` command to configure IP tables, rather calls relevant syscalls in kernel. 

**PROBLEM:**
Descriptor for IP tables rulechain has to be changed as well because it concatenates raw arguments to `iptables` command.

**SOLUTION:**
Add `AppendRules` method to `IPTablesAPI` handler which accepts list of rules and the handler implementation will decide _how_ to append those rules. 

If you are worried about the config option (`minRuleCountForPerfRuleAddition`), you can easily add another method to `IPTablesAPI` that configures this from descriptor and might possibly be ignored by some handler implementations.

P.S. Have you seen this pure Go iptables successor? https://github.com/google/nftables

fgschwan: I agree that this should be hidden, so i changed that in https://github.com/ligato/vpp-agent/pull/1644/commits/21029341e7c21174b8cc86de096a67cdbe653a23 .

However, I didn't wanted to polute API with one method setting `minRuleCountForPerfRuleAddition` so i added config into handler initialization so it can grab it from there. Unfortunately there has been a dependency cycle so i split the config structure into 2 structures to prevent that. Hence the `handler:` in plugin config. Despite the config change, this seem to me cleaner than to change the API as you suggested.

Btw. the [google/nftables](https://github.com/google/nftables) uses iptables successor in linux kernel. That mean other syntax (currently the model of iptablesplugin has raw arguments tied to iptables tool as you mentioned), rename of all "iptables" strings in the code and so on. So switch to nftables will be big change break anyway. It more feels like reimplementing the whole plugin.

PR #1654, opened 2020-04-27
fix: Fix IPSec tun protect + add IPSec e2e test

rastislavszabo:

### DIFF
--- a/tests/e2e/070_ipsec_test.go
+++ b/tests/e2e/070_ipsec_test.go
@@ -117,8 +117,8 @@ func TestIPSec(t *testing.T) {
 	}
 	tpNew := &vpp_ipsec.TunnelProtection{
 		Interface: tunnelIfName,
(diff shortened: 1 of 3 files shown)


tests/e2e/070_ipsec_test.go
```diff
@@ -0,0 +1,166 @@
...
+	"testing"
+
+	. "github.com/onsi/gomega"
+
+	"go.ligato.io/vpp-agent/v3/proto/ligato/kvscheduler"
+	"go.ligato.io/vpp-agent/v3/proto/ligato/vpp/interfaces"
+	"go.ligato.io/vpp-agent/v3/proto/ligato/vpp/ipsec"
+)
+
+func TestSpan(t *testing.T) {
```
milanlenco, line 28: Thanks for test to reproduce the panic.
Only name should be probably something else than `TestSpan`, otherwise good to merge.

PR #1593, opened 2019-12-17
refactor: Improve VPP handler versions and add VPP proxy support

ondrej-fabry:
- [x] add VPP proxy support for remote access to VPP
- [x] improve VPP version resolution and make it more robust
  - compatible version is now determined by checking all messages of one binapi version
- [x] add tests for VPP handlers
- [x] cleanup all VPP handlers for consistency

PR #1677, opened 2020-07-15
refactor: Delete unused REST handler for VPP commands

rewenset:
That handler:
```go
func (p *Plugin) commandHandler(formatter *render.Render) http.HandlerFunc {
```
was moved to govppmux in this PR: https://github.com/ligato/vpp-agent/pull/1593.

And from that point (after the PR was merged), new home for the handler is here:
https://github.com/ligato/vpp-agent/blob/6ba0dbb39e7b96f05fd11caac2df7667b88864a2/plugins/govppmux/rest.go#L58

### DIFF
--- a/plugins/restapi/handlers.go
+++ b/plugins/restapi/handlers.go
@@ -18,9 +18,7 @@ package restapi
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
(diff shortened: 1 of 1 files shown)


rewenset:
(it a screenshot, do not click on links)
image
:open_mouth:

PR #1679, opened 2020-07-29
refactor: separate model for IPSec Security Policies

milanlenco:
The previous model of configuring SPs inside SPD is now deprecated. In practice the content of SPD is changing very frequently and using a single model with all SPs inside is therefore neither convenient nor efficient. A new model SecurityPolicy was added which allows to configure each SP as a separate proto message instance. Although the new model is still backward-compatible with the old one, request to configure SP through the SPD model will return an error (invalid config).

Also fixed couple of bugs that were present in the original implementation and tested the implementation with real-world configurations.


clientv2/linux/dbadapter/data_resync_db.go
```diff
@@ -246,6 +246,12 @@ func (dsl *DataResyncDSL) IPSecSPD(spd *ipsec.SecurityPolicyDatabase) linuxclien
 	return dsl
 }
 
+// IPSecSPD adds Security Policy into the RESYNC request
```
rewenset, line 249: IPSecSP :)

plugins/vpp/ipsecplugin/vppcalls/vpp2001/ipsec_vppcalls.go
```diff
@@ -35,13 +35,13 @@ func (h *IPSecVppHandler) DeleteSPD(spdID uint32) error {
...
+func (h *IPSecVppHandler) AddSP(sp *ipsec.SecurityPolicy) error {
+	return h.spdAddDelEntry(sp, true)
 }
 
 // DeleteSPDEntry implements IPSec handler.
-func (h *IPSecVppHandler) DeleteSPDEntry(spdID, saID uint32, spd *ipsec.SecurityPolicyDatabase_PolicyEntry) error {
-	return h.spdAddDelEntry(spdID, saID, spd, false)
+func (h *IPSecVppHandler) DeleteSP(sp *ipsec.SecurityPolicy) error {
+	return h.spdAddDelEntry(sp, false)
 }
```
rewenset, line 45: two more bad comments with a wrong method name.

PR #1691, opened 2020-07-31
build: Switch to new proto v2

ondrej-fabry:
This PR updates Protobuf-Go to v2

## Protobuf-Go

- Import path changes from `github.com/golang/protobuf` to `google.golang.org/protobuf` 
  (now uses vanity import path and the actual repo is at: https://github.com/protocolbuffers/protobuf-go)

## `v2` is actually `>=v1.20.0`
- The Protobuf-Go repo actually did __not__ increase the major version number! :warning: 
    The project authors decided to not increase major number, because of the import path changing completely
    (which is actually the important part for Go modules when doing breaking changes to an API)
- The new [Protobuf-Go v2](https://pkg.go.dev/mod/google.golang.org/protobuf) has much better API that is easier to work with, providing true [Protobuf reflection](https://pkg.go.dev/google.golang.org/protobuf/reflect/protoreflect?tab=doc#pkg-overview) descriptors (instead of Go type reflection)
- The new interface for `proto.Message` has just a single method that returns the message reflect
- Generated messages no longer export fields/funcs intended for internal use 🎊 (e.g. `XXX_*`)
  - This makes Go docs for generated packages very clean

## `v1` is actually `<=v1.4.0`
The module `github.com/golang/protobuf` has switched its implementation in `v1.4.0` to use the new v2 API under the hood while still providing the legacy v1 API and the v2 API allows backwards compatibility by providing conversions from v1
- `github.com/golang/protobuf v1.4.0` [`go.mod`](https://github.com/golang/protobuf/blob/v1.4.0/go.mod#L7) depends on the new `google.golang.org/protobuf v1.21.0`
- `google.golang.org/protobuf v1.20.0` [`go.mod`](https://github.com/protocolbuffers/protobuf-go/blob/v1.20.0/go.mod#L6) depends on the old `github.com/golang/protobuf v1.4.0-rc.2`

## gRPC
- The gRPC generator is no longer part (plugin) of `protoc-gen-go` and has been moved to [`grpc-go`](https://github.com/grpc/grpc-go) repo and now it is a separate generator `protoc-gen-go-grpc`(executed from the `protoc`)
- gRPC related code is now generated in a separate files - `*_grpc.pb.go`
- Implementators of gRPC server now must embed a generated dummy implementation of gRPC server `Unimplemented*` 
    (this is most likely to avoid panics on when embedding nil interface)

#### Links for more info
- Release notes for v1.4.0: https://github.com/golang/protobuf/releases/tag/v1.4.0
- Read Golang.org blog post for info: https://blog.golang.org/protobuf-apiv2
- Protobuf-Go README for details: https://github.com/protocolbuffers/protobuf-go
- Official Protobuf Docs Go FAQ: https://developers.google.com/protocol-buffers/docs/reference/go/faq


pkg/models/model.go
```diff
@@ -18,8 +18,8 @@ import (
 	"reflect"
 	"strings"
 
-	"github.com/golang/protobuf/descriptor"
 	"github.com/golang/protobuf/proto"
+	"google.golang.org/protobuf/reflect/protoreflect"
```
rewenset, line 22: Is it okay that "proto" package is still imported from `github.com/golang/protobuf/proto` and not from `google.golang.org/protobuf...` ?

ondrej-fabry: Yes, module `github.com/golang/protobuf` has actually switched its implementation in v1.4 to the new API under the hood. See release notes for more info: https://github.com/golang/protobuf/releases/tag/v1.4.0 

You can also see that the `github.com/golang/protobuf` v1.4 (final release for v1) actually has a dependency on the new API `google.golang.org/protobuf` v1.21 which is practically a new major version - v2.

rewenset: Thanks a lot

PR #1703, opened 2020-08-25
feat: Generate graph images using Go library

ondrej-fabry:
- it is no longer needed to install graphviz packge
- graph image output made more compact


ondrej-fabry:
Sample output:

image

milanlenco (approved):

PR #1725, opened 2020-09-10
feat: Changed nat44 pool key to prevent possible key collisions

fgschwan:
I have changed the (etcd/kvstore) key for NAT44 address pool to ensure that there is no key collision in case of 2 configuration that differ in field not included in key (i.e. twice_nat). I have changed the key to the value of a new field called Name (unique key value) so that future field additions couldn't bring this kind of issues again.
To not break existing clients, i support also previous key format, but only the new format doesn't have collision issues. In time, the old format can be removed.

Example of new format (key,value):
`config/vpp/nat/v2/nat44-pool/testpool  {"name":"testpool","vrf_id":0,"first_ip":"169.254.1.2"}`

Example of old format:
`config/vpp/nat/v2/nat44-pool/vrf/0/address/169.254.1.2  {"vrf_id":0,"first_ip":"169.254.1.2"}`

I added also unit test.


ondrej-fabry (approved):

milanlenco:
I see that this is already merged, but could you Filip also please add simple e2e tests here: https://github.com/ligato/vpp-agent/blob/master/tests/e2e/050_nat_test.go
No need to test with traffic between microservices, just some CRUD test with NAT pools to cover the changes in the Retrieve method.
The best way to test CRUD operations is to also include:
```
Expect(ctx.agentInSync()).To(BeTrue(), "Agent is not in-sync")
```
To ensure that Retrieve properly returns what Create/Update/Delete configured.
Thanks

fgschwan:
@milanlenco added tests in new PR https://github.com/ligato/vpp-agent/pull/1725

PR #1726, opened 2020-09-11
feat: add IP protocol number to ACL model

milanlenco:


proto/ligato/vpp/acl/acl.proto
```diff
@@ -41,6 +41,25 @@ message ACL {
...
+                    UDP   = 17;
+                    GRE   = 47;
+                    ESP   = 50;
+                    AH    = 51;
+                    ICMP6 = 58;
+                    EIGRP = 88;
+                    OSPF  = 89;
+                    SCTP  = 132;
+                };
+                Protocol protocol = 3;
```
ondrej-fabry, line 62: I'm not sure this is a good idea.. I just merged a PR where we had to deprecate an enum field because it did not allow users to use custom values. This is because in Protobuf enums cannot have value outside of defined values. 

Perhaps we could keep the enum definition as a helper, but use uint32 for the field? Or something else?

milanlenco: You are the second person to suggest this change and I agree. This was just copied from VPP binary API. I'm not sure if these are the only protocols that VPP can match using ACLs. It would be a strange limitation because the matching algorithm could just compare one integer in IP header without having to understand what protocol it is, but maybe I'm missing something...
I will provide update later today.

ondrej-fabry: Keeping the enum definition might be actually confusing to users I think... best would be to just have `protocol` field as `uint32` type and provide link to the official spec for protocol numbers: http://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml

milanlenco: done

ondrej-fabry (approved):

PR #1731, opened 2020-09-29
Wireguard plugin support

glazychev-art:
The 20.09 release adds support for the wireguard protocol.


plugins/vpp/wireguardplugin/descriptor/peer.go
```diff
@@ -0,0 +1,113 @@
...
+	return oldPeer.PublicKey == newPeer.PublicKey &&
+		oldPeer.Port == newPeer.Port
+}
+
+// Create adds a new wireguard peer.
+func (d *WgPeerDescriptor) Create(key string, peer *wg.Peer) (metadata *wgidx.WgMetadata, err error) {
+	var vppWgIndex uint32
+	vppWgIndex, err = d.wgHandler.AddPeer(peer)
+	if err != nil {
+		d.log.Error(err)
```
ondrej-fabry, line 76: Func should return err here.

plugins/vpp/wireguardplugin/vppcalls/vpp2009/vppcalls_handlers.go
```diff
@@ -0,0 +1,84 @@
...
+	callsChannel govppapi.Channel
+	ifIndexes    ifaceidx.IfaceMetadataIndex
+	log          logging.Logger
+}
+
+func NewWgVppHandler(ch govppapi.Channel, ifIdx ifaceidx.IfaceMetadataIndex, log logging.Logger) vppcalls.WgVppAPI {
+	return &WgVppHandler{ch, ifIdx, log}
+}
+
+func IPToAddress(ipstr string) (addr ip_types.Address, err error) {
```
ondrej-fabry, line 49: There is no need to redefine this conversion function. It's already available in the generated binapi code: https://github.com/ligato/vpp-agent/blob/master/plugins/vpp/binapi/vpp2009/ip_types/ip_types.ba.go#L332

plugins/vpp/wireguardplugin/vppcalls/vpp2009/vppcalls_handlers.go
```diff
@@ -0,0 +1,84 @@
...
+	} else {
+		addr.Af = ip_types.ADDRESS_IP4
+		var ip4addr ip_types.IP4Address
+		copy(ip4addr[:], ip4)
+		addr.Un.SetIP4(ip4addr)
+	}
+	return
+}
+
+func IPToPrefix(prefstr string) (prefix ip_types.Prefix, err error) {
```
ondrej-fabry, line 68: Same here, use the generated conversion func:

https://github.com/ligato/vpp-agent/blob/master/plugins/vpp/binapi/vpp2009/ip_types/ip_types.ba.go#L509

plugins/vpp/wireguardplugin/descriptor/peer.go
```diff
@@ -0,0 +1,113 @@
...
+		WithMetadata:    true,
+		Create:          d.Create,
+		Delete:          d.Delete,
+		Retrieve:        d.Retrieve,
+	}
+}
+
+func (d *WgPeerDescriptor) EquivalentWgPeers(key string, oldPeer, newPeer *wg.Peer) bool {
+	// compare base fields
+	return oldPeer.PublicKey == newPeer.PublicKey &&
```
ondrej-fabry, line 67: Why only comparing `PublicKey` and `Port`? This will cause the changes that modify other fields only to be ignored and not executed at all. If you want to execute (re-create) the object in the VPP you should rather use `proto.Equal` that will compare all the fields for you. The selective comparison is only needed when some values have same meaning even though they have different values - e.g. IP address `0.0.0.0` and _empty string_ is different value but ignored.

[`ligato.vpp.interfaces.WireguardLink`](https://github.com/glazychev-art/vpp-agent/blob/c1b155b9b54873245fed556ea9a4a6b4bb74ac6b/proto/ligato/vpp/interfaces/interface.proto#L387-L404)

go.mod
```diff
@@ -8,6 +8,7 @@ require (
 	github.com/Microsoft/go-winio v0.4.7 // indirect
 	github.com/Nvveen/Gotty v0.0.0-20120604004816-cd527374f1e5 // indirect
 	github.com/Shopify/sarama v1.20.1 // indirect
+	github.com/alecthomas/gometalinter v3.0.0+incompatible // indirect
```
rewenset, line 11: please run `go mod tidy`

plugins/vpp/wireguardplugin/vppcalls/vpp2009/dump_vppcalls.go
```diff
@@ -0,0 +1,95 @@
...
+				ipAddr = fmt.Sprintf("%s/%d", net.IP(ipByte).To16().String(), uint32(prefix.Len))
+			} else {
+				ipAddr = fmt.Sprintf("%s/%d", net.IP(ipByte[:4]).To4().String(), uint32(prefix.Len))
+			}
+			peerDetails.AllowedIps = append(peerDetails.AllowedIps, ipAddr)
+		}
+
+		ifName, _, exists := h.ifIndexes.LookupBySwIfIndex(uint32(vppPeerDetails.Peer.SwIfIndex))
+		if !exists {
+			continue
```
ondrej-fabry, line 73: At least print some debug log here so this case is not hidden from users if they encounter it.

plugins/vpp/wireguardplugin/vppcalls/vpp2009/wireguard_vppcalls.go
```diff
@@ -0,0 +1,87 @@
...
+
+func (h *WgVppHandler) AddPeer(peer *wg.Peer) (uint32, error) {
+	inv_peer_idx := ^uint32(0)
+
+	peer_vpp := vpp_wg.WireguardPeer{
+		Port:                uint16(peer.Port),
+		PersistentKeepalive: uint16(peer.PersistentKeepalive),
+		TableID:             peer.TableId,
+		Flags:               vpp_wg.WireguardPeerFlags(peer.Flags),
+		NAllowedIps:         uint8(peer.NAllowedIp),
```
ondrej-fabry, line 33: There is no need to set value of `NAllowerIps`. The value is ignored and set to the slice length of `AllowedIps` field:

https://github.com/glazychev-art/vpp-agent/blob/wireguard_plugin/plugins/vpp/binapi/vpp2009/wireguard/wireguard.ba.go#L394

plugins/vpp/ifplugin/vppcalls/vpp2009/wireguard_vppcalls.go
```diff
@@ -0,0 +1,119 @@
...
+		stop, err := reqCtx.ReceiveReply(wgDetails)
+		if stop {
+			break // Break from the loop.
+		}
+		if err != nil {
+			return fmt.Errorf("failed to dump Wireguard interface details: %v", err)
+		}
+		_, ifIdxExists := ifc[uint32(wgDetails.Interface.SwIfIndex)]
+		if !ifIdxExists {
+			continue
```
ondrej-fabry, line 97: Add debug log here so we know that wireguard dumped interface unknown index.

plugins/vpp/ifplugin/vppcalls/vpp2009/wireguard_vppcalls.go
```diff
@@ -0,0 +1,119 @@
...
+	"go.ligato.io/vpp-agent/v3/plugins/vpp/binapi/vpp2009/interface_types"
+	"go.ligato.io/vpp-agent/v3/plugins/vpp/binapi/vpp2009/ip_types"
+	"go.ligato.io/vpp-agent/v3/plugins/vpp/binapi/vpp2009/wireguard"
+	"go.ligato.io/vpp-agent/v3/plugins/vpp/ifplugin/vppcalls"
+	interfaces "go.ligato.io/vpp-agent/v3/proto/ligato/vpp/interfaces"
+	"net"
+)
+
+// AddWireguardTunnel adds new Wireguard tunnel interface.
+func (h *InterfaceVppHandler) AddWireguardTunnel(ifName string, wireguardLink *interfaces.WireguardLink) (uint32, error) {
```
ondrej-fabry, line 29: Add checks if wireguard rpc client is nil to avoid sending requests in case the wireguard plugin is disabled. Similar how wmxnet3 interface is done: https://github.com/glazychev-art/vpp-agent/blob/wireguard_plugin/plugins/vpp/ifplugin/vppcalls/vpp2009/vmxnet3_vppcalls.go#L30-L32

proto/ligato/vpp/wireguard/wireguard.proto
```diff
@@ -0,0 +1,35 @@
...
+  string public_key = 1;
+
+  // Peer UDP port
+  uint32 port = 2;
+
+  // Keepalive interval (sec)
+  uint32 persistent_keepalive = 3;
+
+  //  Table ID
+  uint32 table_id = 4;
```
ondrej-fabry, line 18: `Table ID` is most likely referencing to VRF table, so this should be named `vrf_id` to be consistent with other proto models.

proto/ligato/vpp/wireguard/models.go
```diff
@@ -0,0 +1,30 @@
...
+
+// ModuleName is the module name used for models.
+const ModuleName = "vpp.wg"
+
+var (
+	ModelPeer = models.Register(&Peer{}, models.Spec{
+		Module:  ModuleName,
+		Version: "v1",
+		Type:    "peer",
+	}, models.WithNameTemplate("{{.PublicKey}}"))
```
ondrej-fabry, line 29: Is the `PublicKey` enough to uniquely identify different values? Are two wireguard peers with the same  `PublicKey`, but different `vrf_id` or `port` or `endpoint` allowed? If so, these will have collision and might cause ugly bugs.

proto/ligato/vpp/wireguard/wireguard.proto
```diff
@@ -0,0 +1,35 @@
...
+
+  // Endpoint IP
+  string endpoint = 5;
+
+  // Wireguard interface name
+  string interface = 6;
```
ondrej-fabry, line 24: I'm not sure here.. but the name `Interface` might be potentially confusing to users. At first I thought it defines the interface name of the peer.. but I suspect it's the name of the actual interface created by wireguard, right?

[In similar cases](
https://github.com/ligato/vpp-agent/blob/master/proto/ligato/vpp/interfaces/interface.proto#L228) we use `*_if_name` suffix, perhaps `wireguard_if_name` or `wg_if_name` might make it more different to other `interface` / `name` fields that refer to the logical name of interfaces used in the vpp-agent internally, even though it's more "ugly". Perhaps just provide more info in the field comment to make it clear to users what it refers to.

CC @milanlenco

proto/ligato/vpp/interfaces/interface.proto
```diff
@@ -380,4 +382,23 @@ message IPIPLink {
     string src_addr = 2;
     // Destination VTEP IP address
     string dst_addr = 3;
+}
+
+message WireguardLink {
+  uint32 user_instance = 1;
+
+  // Interface index
+  uint32 interface = 2;
```
ondrej-fabry, line 391: We DO NOT store interface index from VPP's runtime in the NB Proto API - the index is by definition only valid in the currently running session and might cause unexpected behaviour after VPP restart.

Change this to string so it refers to logical name of the interface and use the name-index mapping to store the reference between them.

proto/ligato/vpp/interfaces/interface.proto
```diff
@@ -380,4 +382,23 @@ message IPIPLink {
...
+
+  // Generate key automatically
+  bool generate_key = 3;
```
ondrej-fabry, line 394: Note that I might not have enough info to make this assumption, but could this field be possibly unnecessary and we could assume it is true if `private_key` is empty string? Does wireguard interface without private key make any sense? If it does not and must have private key defined or have generate key set to true then it might be simpler to just automatically set key to be generated if it is not set.

CC @milanlenco

proto/ligato/vpp/interfaces/interface.proto
```diff
@@ -380,4 +382,23 @@ message IPIPLink {
     string src_addr = 2;
     // Destination VTEP IP address
     string dst_addr = 3;
+}
+
+message WireguardLink {
+  uint32 user_instance = 1;
```
ondrej-fabry, line 388: What does this refer to? Is this some wireguard internal property?

plugins/vpp/wireguardplugin/descriptor/peer.go
```diff
@@ -0,0 +1,113 @@
...
+		NBKeyPrefix:     wg.ModelPeer.KeyPrefix(),
+		ValueTypeName:   wg.ModelPeer.ProtoName(),
+		KeySelector:     wg.ModelPeer.IsKeyValid,
+		KeyLabel:        wg.ModelPeer.StripKeyPrefix,
+		ValueComparator: d.EquivalentWgPeers,
+		WithMetadata:    true,
+		Create:          d.Create,
+		Delete:          d.Delete,
+		Retrieve:        d.Retrieve,
+	}
```
ondrej-fabry, line 62: Wireguard peer should have dependencies for VRF table if set to non-0 ([example from code](https://github.com/glazychev-art/vpp-agent/blob/wireguard_plugin/plugins/vpp/l3plugin/descriptor/dhcp_proxy.go#L100-L104))

ondrej-fabry, line 62: Add some validation method to validate the fields:
- public_key length
- port value
- persistent_keepalive range
- endpoint IP address format
- wireguard interface length? symbols?
- allowed IPs format
- ...

proto/ligato/vpp/wireguard/wireguard.proto
```diff
@@ -0,0 +1,35 @@
...
+
+  // Flags
+  // WIREGUARD_PEER_STATUS_DEAD = 0x1
+  uint32 flags = 7;
+
+  // Allowed IPs count
+  uint32 n_allowed_ip = 8;
```
milanlenco, line 31: no need for this field - this is not as low-level as VPP binary API, you can do `len(AllowedIps)`

plugins/vpp/wireguardplugin/descriptor/peer.go
```diff
@@ -0,0 +1,113 @@
...
+		Name:            PeerDescriptorName,
+		NBKeyPrefix:     wg.ModelPeer.KeyPrefix(),
+		ValueTypeName:   wg.ModelPeer.ProtoName(),
+		KeySelector:     wg.ModelPeer.IsKeyValid,
+		KeyLabel:        wg.ModelPeer.StripKeyPrefix,
+		ValueComparator: d.EquivalentWgPeers,
+		WithMetadata:    true,
+		Create:          d.Create,
+		Delete:          d.Delete,
+		Retrieve:        d.Retrieve,
```
milanlenco, line 61: Also add here [this line](https://github.com/ligato/vpp-agent/blob/master/plugins/vpp/l2plugin/descriptor/bridgedomain.go#L93)
It is to ensure that interfaces are dumped before WG peers (wgHandler.DumpWgPeers uses interface index to translate index to name and so this index has to be already populated with up-to-date data)

milanlenco (changes requested):

plugins/vpp/wireguardplugin/descriptor/peer.go
glazychev-art: Done

plugins/vpp/wireguardplugin/vppcalls/vpp2009/vppcalls_handlers.go
glazychev-art: Done

plugins/vpp/wireguardplugin/vppcalls/vpp2009/vppcalls_handlers.go
glazychev-art: Done

plugins/vpp/wireguardplugin/descriptor/peer.go
glazychev-art: Yes, it seems `proto.Equal` is better

plugins/vpp/wireguardplugin/vppcalls/vpp2009/dump_vppcalls.go
glazychev-art: Done

proto/ligato/vpp/wireguard/wireguard.proto
glazychev-art: You are right.
Done

proto/ligato/vpp/wireguard/wireguard.proto
```diff
@@ -0,0 +1,33 @@
...
+
+  // Peer UDP port
+  uint32 port = 2;
+
+  // Keepalive interval (sec)
+  uint32 persistent_keepalive = 3;
+
+  // VRF ID
+  // The IP table in which 'endpoint' is reachable
+  uint32 vrf_id = 4;
```
milanlenco, line 19: Isn't VRF given by the tunnel interface?
You can obtain VRF from the interface metadata: https://github.com/ligato/vpp-agent/blob/master/plugins/vpp/ifplugin/ifaceidx/ifaceidx.go#L63

proto/ligato/vpp/interfaces/interface.proto
glazychev-art: Seems like it really doesn't make sense.
Removed `generate_key` field

proto/ligato/vpp/wireguard/models.go
```diff
@@ -0,0 +1,30 @@
...
+
+// ModuleName is the module name used for models.
+const ModuleName = "vpp.wg"
+
+var (
+	ModelPeer = models.Register(&Peer{}, models.Spec{
+		Module:  ModuleName,
+		Version: "v1",
+		Type:    "peer",
+	}, models.WithNameTemplate("{{.WgIfName}}/key/{{.PublicKey}}"))
```
milanlenco, line 29: What about:

{{.WgIfName}}/endpoint/{{.Endpoint}}/{{.Port}}

as a peer identifier? Public key is a bit too long to use as a peer identifier IMHO.

proto/ligato/vpp/wireguard/wireguard.proto
glazychev-art: yep it should

proto/ligato/vpp/wireguard/models.go
glazychev-art: Yes, it seems to be a unique identifier

plugins/vpp/wireguardplugin/vppcalls/vpp2009/dump_vppcalls.go
```diff
@@ -0,0 +1,94 @@
...
+			var ipAddr string
+			ipByte := make([]byte, 16)
+			copy(ipByte[:], prefix.Address.Un.XXX_UnionData[:])
+			if prefix.Address.Af == ip_types.ADDRESS_IP6 {
+				ipAddr = fmt.Sprintf("%s/%d", net.IP(ipByte).To16().String(), uint32(prefix.Len))
+			} else {
+				ipAddr = fmt.Sprintf("%s/%d", net.IP(ipByte[:4]).To4().String(), uint32(prefix.Len))
+			}
+			peerDetails.AllowedIps = append(peerDetails.AllowedIps, ipAddr)
+		}
```
ondrej-fabry, line 67: `len(AllowedIps)` is same as `NAllowedIps` and `ip_types.Prefix` has `String()` method for converting to string so you can simplify this:

```suggestion
for _, prefix := range vppPeerDetails.Peer.AllowedIps {
    peerDetails.AllowedIps = append(peerDetails.AllowedIps, prefix.String()
}
```

plugins/vpp/wireguardplugin/vppcalls/vpp2009/dump_vppcalls.go
```diff
@@ -0,0 +1,94 @@
...
+		peerDetails.WgIfName = ifName;
+
+		var endpointAddr net.IP
+		if vppPeerDetails.Peer.Endpoint.Af == ip_types.ADDRESS_IP6 {
+			src := vppPeerDetails.Peer.Endpoint.Un.GetIP6()
+			endpointAddr = net.IP(src[:])
+		} else {
+			src := vppPeerDetails.Peer.Endpoint.Un.GetIP4()
+			endpointAddr = net.IP(src[:])
+		}
```
ondrej-fabry, line 84:
```suggestion
		endpointAddr := vppPeerDetails.Peer.Endpoint.ToIP()
```

glazychev-art: Yep, you are right.
Thank you!

ondrej-fabry (approved):

ondrej-fabry:
Could you please rebase on master? Support for VPP 19.04 was removed so you can omit it in this PR. Otherwise LGTM

milanlenco (dismissed):

plugins/vpp/wireguardplugin/wireguardplugin.go
```diff
@@ -0,0 +1,72 @@
...
+	IfPlugin    ifplugin.API
+	StatusCheck statuscheck.PluginStatusWriter // optional
+}
+
+func (p *WgPlugin) Init() (err error) {
+	// init Wg handler
+	p.WgHandler = vppcalls.CompatibleWgVppHandler(p.VPP, p.IfPlugin.GetInterfaceIndex(), p.Log)
+	if p.WgHandler == nil {
+		return errors.New("Wireguard handler is not available")
+	}
```
rewenset, line 54: Agent with VPP 20.05 does not want to start.
```
time="2020-10-05 15:05:44.23681" level=debug msg="-> Init(): vpp-wg-plugin" loc="agent/agent.go(220)" logger=agent
time="2020-10-05 15:05:44.23689" level=debug msg="VPP handler wg incompatible with 20.09-rc0 (12 messages)" loc="logging/log_api.go(37)" logger=defaultLogger
time="2020-10-05 15:05:44.23693" level=debug msg="no compatible version found for handler wg: incompatible handler" loc="logging/log_api.go(37)" logger=defaultLogger
time="2020-10-05 15:05:44.23703" level=fatal msg="Wireguard handler is not available" loc="vpp-agent/main.go(87)" logger=defaultLogger
time="2020-10-05 15:05:45.73858" level=info msg="Signal terminated received, stopping." loc="agent/agent.go(196)" logger=agent
```
Maybe log and return no error here?

rewenset: @ondrej-fabry

ondrej-fabry: Add check at the start of `Init()`:
```go
if !p.VPP.IsPluginLoaded("wireguard") {
  p.Log.Warnf("VPP plugin wireguard was disabled by VPP")
  return nil
}
```

similar to NAT plugin.

ondrej-fabry: @glazychev-art could you please fix this along with my [previous comment](https://github.com/ligato/vpp-agent/pull/1731#issuecomment-703579972)?

glazychev-art: @ondrej-fabry Sorry for waiting. I fixed your comments.

ondrej-fabry (approved):

milanlenco (approved):

PR #1735, opened 2020-10-06
feat: VRRP support

NotSoFancyName:
Virtual Router Redundancy Protocol (VRRP)
VPP API: https://git.fd.io/vpp/tree/src/plugins/vrrp/vrrp.api


plugins/vpp/l3plugin/vppcalls/l3_vppcalls.go
```diff
@@ -232,6 +236,15 @@ type TeibVppRead interface {
...
 }
 
+// VrrpVppAPI provides methods for managing VPP VRRP.
+type VrrpVppAPI interface {
+	VppAddVrrp(entry *l3.VRRPEntry) error
+	VppDelVrrp(entry *l3.VRRPEntry) error
+	VppStartVrrp(entry *l3.VRRPEntry) error
+	VppStopVrrp(entry *l3.VRRPEntry) error
+	DumpVrrpEntries() ([]*l3.VRRPEntry, error)
+}
```
NotSoFancyName, line 246: @milanlenco Could you tell me whether other [API](https://docs.fd.io/vpp/21.01/d3/dbe/vrrp_8api.html) methods should be implemented (peer and track_if) or these are sufficient for now?

plugins/vpp/l3plugin/descriptor/vrrp.go
```diff
@@ -0,0 +1,235 @@
...
+		})
+	}
+
+	return retrieved, nil
+}
+
+// IsRetriableFailure returns false if error is one of errors
+// defined at the top of this file as non-retriable.
+func (d *VrrpDescriptor) IsRetriableFailure(err error) bool {
+	if errors.Is(err, vppcalls.ErrVRRPUnsupported) {
```
ondrej-fabry, line 231: You could register this error as global unretrieble error: https://github.com/ligato/vpp-agent/blob/master/plugins/kvscheduler/descriptor_handler.go#L22

plugins/vpp/l3plugin/vppcalls/vpp2005/vrrp_dump.go
```diff
@@ -0,0 +1,93 @@
...
+			break
+		}
+		if err != nil {
+			return nil, err
+		}
+
+		// VRRP interface
+		ifName, _, exists := h.ifIndexes.LookupBySwIfIndex(uint32(vrrpDetails.Config.SwIfIndex))
+		if !exists {
+			h.log.Warnf("VRRP dump: interface name not found for index %d", vrrpDetails.Config.SwIfIndex)
```
ondrej-fabry, line 44: This will store `Interface` field as empty even though it refers to some unknown interface.. not sure what would be the right way to process such data.. @milanlenco  thoughts?

proto/ligato/vpp/l3/vrrp.proto
```diff
@@ -0,0 +1,29 @@
...
+    // Controls whether a virtual router in Master state will accept packets
+    // addressed to the address owner's IPvX address as its own if it is not the IPvX address owner.
+    bool accept = 6;
+    // Unicast mode may be used to take
+    // advantage of newer token ring adapter implementations that support
+    // non-promiscuous reception for multiple unicast MAC addresses and to
+    // avoid both the multicast traffic and usage conflicts associated with
+    // the use of token ring functional addresses.
+    bool unicast = 7;
+    bool ipv6 = 8;
```
ondrej-fabry, line 25: Isn't this implicitly given by the `addrs` ?

proto/ligato/vpp/l3/vrrp.proto
```diff
@@ -0,0 +1,29 @@
...
+package ligato.vpp.l3;
+
+option go_package = "go.ligato.io/vpp-agent/v3/proto/ligato/vpp/l3;vpp_l3";
+
+message VRRPEntry {
+    string interface = 1;
+    uint32 vr_id = 2;
+    uint32 priority = 3;
+    // VR advertisement interval.
+    uint32 interval = 4;
```
ondrej-fabry, line 12: What units does this interval use?

proto/ligato/vpp/l3/vrrp.proto
```diff
@@ -0,0 +1,29 @@
...
+    bool accept = 6;
+    // Unicast mode may be used to take
+    // advantage of newer token ring adapter implementations that support
+    // non-promiscuous reception for multiple unicast MAC addresses and to
+    // avoid both the multicast traffic and usage conflicts associated with
+    // the use of token ring functional addresses.
+    bool unicast = 7;
+    bool ipv6 = 8;
+
+    repeated string addrs = 9;
```
ondrej-fabry, line 27: If these are IP addresses it would be better named `ip_adresses` to be more consistent and clear to user.

proto/ligato/vpp/l3/vrrp.proto
```diff
@@ -0,0 +1,29 @@
+syntax = "proto3";
+
+package ligato.vpp.l3;
+
+option go_package = "go.ligato.io/vpp-agent/v3/proto/ligato/vpp/l3;vpp_l3";
+
+message VRRPEntry {
```
ondrej-fabry, line 7: Perhaps, this message could at least explain the abbreviation VRRP.

plugins/vpp/l3plugin/vppcalls/vpp2009/vrrp_vppcalls.go
```diff
@@ -0,0 +1,119 @@
...
+// VppDelVrrp implements VRRP handler.
+func (h *VrrpVppHandler) VppDelVrrp(entry *l3.VRRPEntry) error {
+	return h.vppAddDelVrrp(entry, 0)
+}
+
+func (h *VrrpVppHandler) vppStartStopVrrp(entry *l3.VRRPEntry, isStart uint8) error {
+
+	md, exist := h.ifIndexes.LookupByName(entry.Interface)
+	if !exist {
+		return errInvalidInterface
```
ondrej-fabry, line 93: This error should wrap the error with info about what interface it was looking up.

plugins/vpp/l3plugin/vppcalls/vpp2009/vrrp_vppcalls.go
```diff
@@ -0,0 +1,119 @@
...
+		if err != nil {
+			return err
+		}
+
+		addrs = append(addrs, ip)
+	}
+
+	md, exist := h.ifIndexes.LookupByName(entry.Interface)
+	if !exist {
+		return errInvalidInterface
```
ondrej-fabry, line 43: Same case here.

plugins/vpp/l3plugin/descriptor/vrrp.go
```diff
@@ -0,0 +1,235 @@
...
+// Update updates VPP VRRP entry.
+func (d *VrrpDescriptor) Update(_ string, oldVRRPEntry, newVRRPEntry *l3.VRRPEntry, _ interface{}) (
+	_ interface{}, err error) {
+
+	if newVRRPEntry.Enabled {
+		err = d.vrrpHandler.VppStartVrrp(newVRRPEntry)
+	} else {
+		err = d.vrrpHandler.VppStopVrrp(newVRRPEntry)
+	}
+	d.log.Debugf("Unsupported action: ", err)
```
ondrej-fabry, line 193: Log out of place?

plugins/vpp/l3plugin/descriptor/vrrp.go
```diff
@@ -0,0 +1,235 @@
...
+
+	if oldVRRPEntry.Interface == newVRRPEntry.Interface &&
+		oldVRRPEntry.Interval == newVRRPEntry.Interval &&
+		oldVRRPEntry.Priority == newVRRPEntry.Priority &&
+		oldVRRPEntry.VrId == newVRRPEntry.VrId &&
+		oldVRRPEntry.Ipv6 == newVRRPEntry.Ipv6 &&
+		oldVRRPEntry.Accept == newVRRPEntry.Accept &&
+		oldVRRPEntry.Preempt == newVRRPEntry.Preempt &&
+		oldVRRPEntry.Unicast == newVRRPEntry.Unicast &&
+		len(oldVRRPEntry.Addrs) == len(newVRRPEntry.Addrs) {
```
ondrej-fabry, line 171: What if some addresses changes, but overall count stays same? This `if` will return false before the next check that compares the addresses..

plugins/vpp/l3plugin/descriptor/vrrp.go
```diff
@@ -0,0 +1,235 @@
...
+		}
+	}
+	return nil
+}
+
+// Create adds VPP VRRP entry.
+func (d *VrrpDescriptor) Create(key string, vrrp *l3.VRRPEntry) (interface{}, error) {
+	if err := d.vrrpHandler.VppAddVrrp(vrrp); err != nil {
+		if errors.Is(vppcalls.ErrVRRPUnsupported, err) {
+			d.log.Debugf("Unsupported action: ", err)
```
ondrej-fabry, line 127: Debug**f** without formatting directive..

proto/ligato/vpp/l3/vrrp.proto
```diff
@@ -4,11 +4,13 @@ package ligato.vpp.l3;
...
 option go_package = "go.ligato.io/vpp-agent/v3/proto/ligato/vpp/l3;vpp_l3";
 
+// VRRPEntry represents Virtual Router desired state.
 message VRRPEntry {
     string interface = 1;
     uint32 vr_id = 2;
+    // Priority defines which router becomes master.
     uint32 priority = 3;
-    // VR advertisement interval.
+    // VR advertisement interval in centiseconds.
```
ondrej-fabry, line 13: Oh my, I dont believe I actually measured time in centiseconds like ever. I think we should choose something more common, if seconds are too big than just go with milliseconds and convert to "centiseconds" down the line if that's what VPP expects.

NotSoFancyName: Good point.

milanlenco: This strange time unit is used in the RFC for VRRPv3 (VRRPv2 uses seconds): https://tools.ietf.org/html/rfc5798
It is probably the smallest time unit that a VRRP router must support.

plugins/vpp/l3plugin/descriptor/vrrp.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+	return nil, nil
+}
+
+// Delete removes VPP VRRP entry.
+func (d *VrrpDescriptor) Delete(key string, vrrp *l3.VRRPEntry, metadata interface{}) error {
+	if vrrp.Enabled {
+		if err := d.vrrpHandler.VppStopVrrp(vrrp); err != nil {
+			if errors.Is(vppcalls.ErrVRRPUnsupported, err) {
+				d.log.Debugf("Unsupported action: ", err)
```
ondrej-fabry, line 151: Debugf without formatting directive. Use linter in your IDE to find these.

proto/ligato/vpp/l3/vrrp.proto
NotSoFancyName: What is the best way to handle this situation? I am planing to make user put milliseconds and then covert to centiseconds, but some error will be present (for example user puts 55 and gets 50). Should I leave the centiseconds in this case?

ondrej-fabry: I would go for milliseconds, it's definitely more common and users wont need to think much about conversion from centiseconds. The small error wont matter for sure.

milanlenco: I'm OK with milliseconds - maybe just log warning if (value % 10 != 0)

NotSoFancyName: Okay, then I will do the millisecond with option.

plugins/vpp/l3plugin/vppcalls/vpp2009/vrrp_dump.go
```diff
@@ -0,0 +1,89 @@
...
+		}
+		if uintToBool(uint8(vrrpDetails.Config.Flags & vrrp.VRRP_API_VR_PREEMPT)) {
+			isPreempt = true
+		}
+		if uintToBool(uint8(vrrpDetails.Config.Flags & vrrp.VRRP_API_VR_ACCEPT)) {
+			isAccept = true
+		}
+		if uintToBool(uint8(vrrpDetails.Config.Flags & vrrp.VRRP_API_VR_UNICAST)) {
+			isUnicast = true
+		}
```
ondrej-fabry, line 59: This will avoid extra conversions (Flags is uint32 anyway):

```suggestion
			isPreempt = (vrrpDetails.Config.Flags & vrrp.VRRP_API_VR_PREEMPT) == vrrp.VRRP_API_VR_PREEMPT
			isAccept = (vrrpDetails.Config.Flags & vrrp.VRRP_API_VR_ACCEPT) == vrrp.VRRP_API_VR_ACCEPT
			isUnicast = (vrrpDetails.Config.Flags & vrrp.VRRP_API_VR_UNICAST) == vrrp.VRRP_API_VR_UNICAST 
```

ondrej-fabry:
Please rebase on master, there are conflicts.

plugins/vpp/l3plugin/descriptor/vrrp.go
```diff
@@ -0,0 +1,255 @@
...
+const (
+	// VrrpDescriptorName is the name of the descriptor.
+	VrrpDescriptorName = "vrrp"
+
+	// Dependency labels.
+	vrrpEntryInterfaceDep    = "interface-exists"
+	vrrpDescriptorLoggerName = "vrrp-descriptor"
+
+	maxUint8  = 255
+	maxUint16 = 65535
```
ondrej-fabry, line 40: These are defined in the [`math`](https://golang.org/pkg/math/#pkg-constants) package already.

ondrej-fabry (dismissed):

plugins/vpp/l3plugin/descriptor/vrrp.go
```diff
@@ -0,0 +1,254 @@
...
+func (d *VrrpDescriptor) Retrieve(correlate []adapter.VRRPEntryKVWithMetadata) (
+	retrieved []adapter.VRRPEntryKVWithMetadata, err error,
+) {
+	entries, err := d.vrrpHandler.DumpVrrpEntries()
+
+	for _, entry := range entries {
+		retrieved = append(retrieved, adapter.VRRPEntryKVWithMetadata{
+			Key:    l3.VrrpEntryKey(entry.Interface, entry.VrId),
+			Value:  entry,
+			Origin: kvs.UnknownOrigin,
```
milanlenco, line 209: Here you can do `Origin: kvs.FromNB`
Unknown is used when VPP creates some more items automatically on its own and it is impossible to distinguish between them and items configured by descriptor. For example, VPP populates routing table with some special routes already when interface with IP address is created (even if no routes are configured explicitly).
But VRRP I assume never configures anything implicitly.

milanlenco:
Added just one comment, otherwise good to merge.

plugins/vpp/l3plugin/vppcalls/vpp2005/vrrp_dump.go
```diff
@@ -0,0 +1,84 @@
...
+
+		ipStrs := make([]string, 0, len(vrrpDetails.Addrs))
+		for _, v := range vrrpDetails.Addrs {
+			if v.Af == ip_types.ADDRESS_IP4 {
+				addr := v.Un.GetIP4()
+				ipStrs = append(ipStrs, net.IP(addr[:]).To4().String())
+			} else {
+				addr := v.Un.GetIP6()
+				ipStrs = append(ipStrs, net.IP(addr[:]).To16().String())
+			}
```
ondrej-fabry, line 64: The [`ip_types.Address`](https://github.com/ligato/vpp-agent/blob/master/plugins/vpp/binapi/vpp2009/ip_types/ip_types.ba.go#L327) used in the [`Addrs`](https://github.com/NotSoFancyName/vpp-agent/blob/vrrp-support/plugins/vpp/binapi/vpp2009/vrrp/vrrp.ba.go#L166) field actually has [`String()`](https://github.com/ligato/vpp-agent/blob/master/plugins/vpp/binapi/vpp2009/ip_types/ip_types.ba.go#L360) method. So you can avoid boilerplate:

```suggestion
			ipStrs = append(ipStrs, v.String())
```

plugins/vpp/l3plugin/vppcalls/vpp2005/vrrp_dump.go
milanlenco: I think this is OK because as a result the VRRP entry will be determined as not being in-sync and Update/Re-create should then take care of that.

plugins/vpp/l3plugin/vppcalls/vpp2005/vrrp_vppcalls.go
```diff
@@ -0,0 +1,136 @@
...
+
+const (
+	centiMilliRatio uint32 = 10
+)
+
+func (h *VrrpVppHandler) vppAddDelVrrp(entry *l3.VRRPEntry, isAdd uint8) error {
+	var addrs []ip_types.Address
+	var isIpv6 bool
+	for idx, addr := range entry.IpAddresses {
+		ip, err := ipToAddress(addr)
```
ondrej-fabry, line 34: Even though most of the current code in vppcalls still uses this helper func `ipToAddress`, for new development it's better to use recently added [`ParseAddress()`](https://github.com/ligato/vpp-agent/blob/master/plugins/vpp/binapi/vpp2009/ip_types/ip_types.ba.go#L332) to convert IP from string form to [`ip_types.Address`](https://github.com/ligato/vpp-agent/blob/master/plugins/vpp/binapi/vpp2009/ip_types/ip_types.ba.go#L327) type. The old usage can be switched to this in separate PR.

proto/ligato/vpp/l3/vrrp.proto
```diff
@@ -0,0 +1,31 @@
+syntax = "proto3";
+
+package ligato.vpp.l3;
+
+option go_package = "go.ligato.io/vpp-agent/v3/proto/ligato/vpp/l3;vpp_l3";
+
+// VRRPEntry represents Virtual Router desired state.
+message VRRPEntry {
+    string interface = 1;
+    uint32 vr_id = 2;
```
ondrej-fabry, line 10: Document the maximum allowed value 255.

proto/ligato/vpp/l3/vrrp.proto
```diff
@@ -0,0 +1,31 @@
+syntax = "proto3";
+
+package ligato.vpp.l3;
+
+option go_package = "go.ligato.io/vpp-agent/v3/proto/ligato/vpp/l3;vpp_l3";
+
+// VRRPEntry represents Virtual Router desired state.
+message VRRPEntry {
+    string interface = 1;
```
ondrej-fabry, line 9: Document this field with info about reference to logical interface name.

plugins/vpp/l3plugin/vppcalls/l3_vppcalls.go
```diff
@@ -232,6 +236,15 @@ type TeibVppRead interface {
 	DumpTeib() ([]*l3.TeibEntry, error)
 }
 
+// VrrpVppAPI provides methods for managing VPP VRRP.
+type VrrpVppAPI interface {
+	VppAddVrrp(entry *l3.VRRPEntry) error
+	VppDelVrrp(entry *l3.VRRPEntry) error
+	VppStartVrrp(entry *l3.VRRPEntry) error
+	VppStopVrrp(entry *l3.VRRPEntry) error
+	DumpVrrpEntries() ([]*l3.VRRPEntry, error)
```
ondrej-fabry, line 245: It would be better for `DumpVrrpEntries` to return `VrrpDetails` type that would have field `VRRPEntry *l3.VRRPEntry` and `Meta *VRRPEntryMeta`, so any extra data from [`vrrp.VrrpVrDetails`](https://github.com/NotSoFancyName/vpp-agent/blob/vrrp-support/plugins/vpp/binapi/vpp2009/vrrp/vrrp.ba.go#L271) message can be stored in there. For example it might be useful to store data from [`vrrp.VrrpVrRuntime`](https://github.com/NotSoFancyName/vpp-agent/blob/vrrp-support/plugins/vpp/binapi/vpp2009/vrrp/vrrp.ba.go#L136-L142).

plugins/vpp/l3plugin/vppcalls/vpp2005/vrrp_dump.go
NotSoFancyName: My bad, I have not noticed first, it is present in 2009 but not in 2005, so I guess I will leave it this way in 2005 but change in 2009

plugins/vpp/l3plugin/vppcalls/l3_vppcalls.go
NotSoFancyName: Added rest /dump/vpp/v2/vrrps endpoint for vrrp dump

plugins/vpp/l3plugin/vppcalls/vpp2005/vrrp_vppcalls.go
```diff
@@ -31,6 +31,9 @@ func (h *VrrpVppHandler) vppAddDelVrrp(entry *l3.VRRPEntry, isAdd uint8) error {
 	var addrs []ip_types.Address
 	var isIpv6 bool
 	for idx, addr := range entry.IpAddresses {
+		if addr == "localhost" {
```
ondrej-fabry, line 34: Why handling this specific case? This should actually never happen because it won't pass through Validate: https://github.com/NotSoFancyName/vpp-agent/blob/vrrp-support/plugins/vpp/l3plugin/descriptor/vrrp.go#L115-L118

ondrej-fabry: https://play.golang.org/p/Lv9iWoaTuey

ondrej-fabry (approved):

PR #1744, opened 2020-10-14
feat: Add support for Linux VRFs

milanlenco:
This PR contains these changes in proto:
- adds new special type of linux interface `VRF_DEVICE`
- adds new field `VrfMasterInterface` to linux interface

Documentation for Linux VRF:
- https://www.kernel.org/doc/Documentation/networking/vrf.txt
- https://cumulusnetworks.com/blog/vrf-for-linux/


plugins/linux/ifplugin/descriptor/interface_vrfdev.go
```diff
@@ -0,0 +1,107 @@
...
+		d.log.Error(err)
+		return err
+	}
+	return nil
+}
+
+func (d *InterfaceDescriptor) enableL3MasterDev() error {
+	value, err := getSysctl(sysctlL3MDevVar)
+	if err != nil {
+		err = fmt.Errorf("could not read sysctl value for %s: %v",
```
ondrej-fabry, line 95: It is better to use the format specifier with `%w` for `error` values, so they can be unwrapped by the caller(s). 

quoted from [`fmt.Errorf`](https://golang.org/pkg/fmt/#Errorf) docs:
> If the format specifier includes a %w verb with an error operand, the returned error will implement an Unwrap method returning the operand. It is invalid to include more than one %w verb or to supply it with an operand that does not implement the error interface. The %w verb is otherwise a synonym for %v.

plugins/linux/ifplugin/ifaceidx/ifaceidx.go
```diff
@@ -68,6 +68,8 @@ type LinuxIfMetadata struct {
 	VPPTapName   string // empty for VETHs
 	Namespace    *linux_namespace.NetNamespace
 	HostIfName   string
+	Vrf          string
+	VrfDevRT     int // only set for VRF_DEVICE
```
ondrej-fabry, line 72: Shouldn't this be `uint32` or at least `uint`?

plugins/linux/ifplugin/descriptor/interface_vrf.go
```diff
@@ -0,0 +1,163 @@
...
+	return nil
+}
+
+// Create puts interface into a VRF.
+func (d *InterfaceVrfDescriptor) Create(key string, emptyVal proto.Message) (metadata kvs.Metadata, err error) {
+	iface, vrf, _, _ := interfaces.ParseInterfaceVrfKey(key)
+
+	ifMeta, found := d.intfIndex.LookupByName(iface)
+	if !found {
+		err = errors.Errorf("failed to find interface %s", iface)
```
ondrej-fabry, line 92: Forgot to wrap error here.

plugins/linux/ifplugin/ifaceidx/ifaceidx.go
milanlenco: Agree that it should be and will change it.
I used `int` only because netlink package also uses it for the table ID, but negative values would not be valid, so...

proto/ligato/linux/interfaces/interface.proto
```diff
@@ -52,11 +58,19 @@ message Interface {
...
     };
 
     // Configure/Resync link only. IP/MAC addresses are expected to be configured
     // externally - i.e. by a different agent or manually via CLI.
     bool link_only = 9;
+
+    // VRF to which this interface should be assigned to.
+    // The field should reference the logical name of a VRF device (i.e. instance of this Interface model with
+    // VRF_DEVICE type).
+    string vrf = 10;
```
ondrej-fabry, line 73: I'm not sure about the naming `vrf` here.. we use `*_interface`/`*interfaces` practically everywhere else where interface is references. Using `vrf_interface` is more consistent and possibly clearer to the users.

plugins/linux/ifplugin/ifaceidx/ifaceidx.go
ondrej-fabry: Right. I see they use `int`, but only in [`Route`](https://pkg.go.dev/github.com/vishvananda/netlink#Route). In the [`Vrf`](https://pkg.go.dev/github.com/vishvananda/netlink#Vrf) it uses `uint32`.

plugins/linux/ifplugin/linuxcalls/link_linuxcalls.go
```diff
@@ -232,3 +232,45 @@ func (h *NetLinkHandler) AddVethInterfacePair(ifName, peerIfName string) error {
...
+// AddVRFDevice configures new VRF network device.
+func (h *NetLinkHandler) AddVRFDevice(vrfDevName string, routingTable uint32) error {
+	attrs := netlink.NewLinkAttrs()
+	attrs.Name = vrfDevName
+	link := &netlink.Vrf{
+		LinkAttrs: attrs,
+		Table:     routingTable,
+	}
+	if err := netlink.LinkAdd(link); err != nil {
+		return errors.Wrapf(err, "LinkAdd %v", link)
```
ondrej-fabry, line 245: This error will be printed as:
`failed to add VRF device vrf-1 (rt: 1): LinkAdd &{{0 0 -1 vrf-1  0 0 0 0 <nil>  <nil> 0 <nil>  <nil> unknown 0 0 0 []} 1}: operation not supported`

and adding variable `link` with `%v` verb does not seem very helpful.. since we build the link/attrs it should be enough to mention `vrfDevName` (`attrs.Name`) and `routingTable` (`link.Table`).

proto/ligato/linux/interfaces/interface.proto
milanlenco: What about `vrf_master_interface` or shorter variant `vrf_master_if`?
Just to make it more clear that what is being referenced is not some random interface inside a VRF, but the interface implementing VRF as a whole.
(master/slave terminology is not very popular these days but that's how they call it in Linux - `ip link set dev eth1 master vrf-blue` is said to `enslave` `eth1` under `vrf-blue` interface/device)

ondrej-fabry: Yea, I think that's okay.

milanlenco:
PR updated after the review.

ondrej-fabry (approved):

plugins/linux/ifplugin/linuxcalls/link_linuxcalls.go
```diff
@@ -232,3 +232,49 @@ func (h *NetLinkHandler) AddVethInterfacePair(ifName, peerIfName string) error {
...
+// AddVRFDevice configures new VRF network device.
+func (h *NetLinkHandler) AddVRFDevice(vrfDevName string, routingTable uint32) error {
+	attrs := netlink.NewLinkAttrs()
+	attrs.Name = vrfDevName
+	link := &netlink.Vrf{
+		LinkAttrs: attrs,
+		Table:     routingTable,
+	}
+	if err := netlink.LinkAdd(link); err != nil {
+		return errors.Wrapf(err, "failed to add VRF device: LinkAdd (vrf=%s, rt=%d)",
```
ondrej-fabry, line 245: The part with `failed to add VRF device:` should not be added here, because it is already [added by the caller](https://github.com/milanlenco/vpp-agent/blob/1a3d4fc48832cb02dc3ad3bc47abe5b024349155/plugins/linux/ifplugin/descriptor/interface_vrfdev.go#L55-L59), so it is redundant and would make the error look like:

```
...: creating X interface failed: failed to add VRF device X (rt: Z): failed to add VRF device: LinkAdd (vrf=Y, rt=Z)
```


The errors returned to caller should only add error context that caller does not know about. In this case "failed to **add VRF device**" is not needed because caller knows it since it called `AddVRFDevice` function. What it does not know that this function failed during call to `LinkAdd` so that part is adding useful context info to the error.

ondrej-fabry (approved):

PR #1756, opened 2020-10-28
feat: Implement WaitDone for Delete in Configurator service

### LINKED ISSUE #1750: Need WaitDone flag on Delete operation
The WaitDone flag on Update is awesome, but its also needed for Delete.

ondrej-fabry:
Fixes #1750

- add 'agenctl config delete' command


cmd/agentctl/commands/config.go
```diff
@@ -197,6 +198,101 @@ func runConfigUpdate(cli agentcli.Cli, opts ConfigUpdateOptions, args []string)
...
+		logrus.Debugf("grpc header: %+v", header)
+		if seqNum, ok := header["seqnum"]; ok {
+			ref, _ := strconv.Atoi(seqNum[0])
+			txns, err := cli.Client().SchedulerHistory(ctx, types.SchedulerHistoryOptions{
+				SeqNum: ref,
+			})
+			if err != nil {
+				logrus.Warnf("getting history for seqNum %d failed: %v", ref, err)
+			} else {
+				data = txns
```
VladoLavor, line 280: Don't you lose err/data value from the `client.Delete` here?

ondrej-fabry: This case will only run when `verbose mode is enabled` and `seqnum` is in the grpc header. This means that transactions has been executed and the output from that will be printed, basically showing entire transaction log like in the agent output, which usually contains the error that occurred, normally as the last step.

PR #1757, opened 2020-10-28
Improve testing process for e2e/integration tests

ondrej-fabry:
- fixes linked glibc library for Ubuntu 20.04 (switch to static compilation)
- exports funcs and methods in e2e package (library users can write e2e tests)
- build testing docker images for running tests (easier repeatability)
- kvscheduler graph generation now renders images by running dot program (if cgo disabled)
- gotestsum is now used for running tests (provides better output and summary)


tests/e2e/main_test.go
```diff
@@ -41,6 +41,10 @@ func TestMain(m *testing.M) {
 	if *debug {
 		govppcore.SetLogLevel(logrus.DebugLevel)
 	}
+	if os.Getenv("TRAVIS") != "" {
+		log.Println("skipping e2e tests for Travis")
```
fgschwan, line 45: E2E tests are running on travis right now. I think that only when something is merged master, but running. Why do you want to turn it off? The "TRAVIS" env variable seems like somethings that is always defined (and nonempty) on travis machines. If this variable should switch the tests on or off, would not it be better to use another env variable? Something like "dont-run-e2e-tests" or "e2e-test"="on"/"off".

fgschwan (changes requested):

milanlenco (approved):

tests/e2e/main_test.go
ondrej-fabry: You are totally right. The `TRAVIS` variable is set for the VM on Travis but we run integration/e2e tests in docker images where it is not present and this was quick hack we used to prevent Travis from running e2e/integration tests when executing `go test ./...`.

I have pushed a commit where are I improved this so it's not hardcoded to TRAVIS and requires setting `TEST_SUITE=xxx` to run tests, or additionally possible to enable via build tag `go test -tags e2e ./tests/e2e`.

fgschwan (approved):

PR #1764, opened 2020-11-27
feat: proxied models (remotely learned models used in DefaultRegistry)

milanlenco:
This small PR allows to register models learned remotely (i.e. over gRCP using the meta service) into the LocalRegistry.

Why is this useful?
This change can be used as a building block for integration of multiple (distributed) CNFs (pairs of agent + some data-plane component) into one appliance with a single API endpoint. A central agent can dynamically discover all CNFs, learn their models over gRPC and generate kv descriptors for them that would mostly just forward CRUD operations via remoteclient.
So basically the change allows the following configuration sequence:
```
agentctl/etcd/... -> [proxy agent(s) ->]  target agent
```

To proxy a given model one would write something like:
```
client, err = remoteclient.NewClientGRPC(grpcConn, remoteclient.UseRemoteRegistry("config"))
models, err := swMod.cfgClient.KnownModels("config")
for _, model := range models {
    if !proxyThisModel {
      continue
    }
    spec := models.ToSpec(cnfModel.info.Spec)
    _, err := models.DefaultRegistry.Register(cnfModel.info, spec)
    // build descriptor for it, etc...
}
```

### DIFF
--- a/pkg/models/item.go
+++ b/pkg/models/item.go
@@ -75,17 +75,17 @@ func UnmarshalItem(item *api.Item) (proto.Message, error) {
 // UnmarshalItemUsingModelRegistry is helper function for unmarshalling items (using given model registry)
 func UnmarshalItemUsingModelRegistry(item *api.Item, modelRegistry Registry) (proto.Message, error) {
 	// check existence of known model
-	_, err := GetModelFromModelRegistryForItem(item, modelRegistry)
+	model, err := GetModelFromModelRegistryForItem(item, modelRegistry)
 	if err != nil {
 		return nil, err
 	}
 
 	// unmarshal item's inner data
-	// (we must distinguish between model registries due to different go types produced by using different
-	// model registry. The LocalRegistry use cases need go types as generated from models, but
-	// the RemoteRegistry can't produce such go typed instances (we know the name of go type, but can't
-	// produce it from remote information) so dynamic proto message must be enough (*dynamicpb.Message))
-	if _, ok := modelRegistry.(*LocalRegistry); ok {
+	// We must distinguish between locally and remotely known models with respect to the underlying go type.
+	// LocallyKnownModel is used for proto message with go type generated from imported proto file and known
+	// at compile time, while RemotelyKnownModel can't produce such go typed instances (we know the name of go type,
+	// but can't produce it from remote information) so dynamic proto message must be enough (*dynamicpb.Message))
+	if _, ok := model.(*LocallyKnownModel); ok {
 		return unmarshalItemDataAnyOfLocalModel(item.GetData().GetAny())
 	}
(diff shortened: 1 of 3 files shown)


pkg/models/registry.go
```diff
@@ -96,14 +108,28 @@ func (r *LocalRegistry) MessageTypeRegistry() *protoregistry.Types {
...
+	})
 	return typeRegistry
 }
 
 // Register registers a protobuf message with given model specification.
 // If spec.Class is unset empty it defaults to 'config'.
 func (r *LocalRegistry) Register(x interface{}, spec Spec, opts ...ModelOption) (KnownModel, error) {
-	goType := reflect.TypeOf(x)
+	// check if the model was learned remotely
+	if modelInfo, isProxied := x.(*ModelInfo); isProxied {
```
fgschwan, line 123: The syntax of the "x" paramater doesn't inforce it, but the documentation says that the protobuf message is expected. So i would prefer a proto message as input. In your case you can use this to create new empty proto message:
`dynamicpb.NewMessageType(modelinfo.MessageDescriptor).New().Interface()`
(maybe you can put it somewhere as conversion helper function)
This will create a dynamicpb.Message instance and you can use that type here as type check switch.

fgschwan (changes requested):

pkg/models/registry.go
milanlenco: I can update the function documentation, but I don't think that I can use dynamicpb here since RemoteRegistry requires ModelInfo.

fgschwan: Unfortunately true. So please just update to docs (the registry interface docs probably also need some update).

pkg/models/item.go
```diff
@@ -81,11 +81,15 @@ func UnmarshalItemUsingModelRegistry(item *api.Item, modelRegistry Registry) (pr
...
-	// (we must distinguish between model registries due to different go types produced by using different
-	// model registry. The LocalRegistry use cases need go types as generated from models, but
-	// the RemoteRegistry can't produce such go typed instances (we know the name of go type, but can't
-	// produce it from remote information) so dynamic proto message must be enough (*dynamicpb.Message))
-	if _, ok := modelRegistry.(*LocalRegistry); ok {
+	// We must distinguish between locally and remotely known models with respect to the underlying go type.
+	// LocallyKnownModel is used for proto message with go type generated from imported proto file and known
+	// at compile time, while RemotelyKnownModel can't produce such go typed instances (we know the name of go type,
+	// but can't produce it from remote information) so dynamic proto message must be enough (*dynamicpb.Message))
+	model, err := GetModelFromModelRegistryForItem(item, modelRegistry)
```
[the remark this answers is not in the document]
milanlenco: done

pkg/models/registry.go
milanlenco: done

fgschwan (approved):

ondrej-fabry (approved):

Issue #1569, opened 2019-11-21
Context parameter for KV descriptors


ondrej-fabry:
Use `context.Context` instead of `key string` as first parameter in KV descriptor methods.

This will allow using context up to govpp calls and provide a lot of insight for debugging and future optimization.

milanlenco:
Will start working on this in parallel with my main tasks since the need to have some context in descriptor just keeps reappearing :)

PR #1768, opened 2020-12-04
feat: existing Linux VRFs

milanlenco:
This pull requests adds support for the following use-cases related to Linux interfaces and VRFs:
- Managed interface inside existing (i.e. not managed by the agent) VRF 
- Existing interface inside existing VRF (i.e. agent only waits for this to be configured externally)

These use-cases are very specific but needed for one of our internal projects.

This PR also fixes the issue with '"flapping" link notifications from netlink - i.e. receiving link DOWN for a very short time (<1ms) because something is being configured on the interface. This causes unnecessary re-creation of everything that depends on it.
The solution is to simply delay DOWN notification (unless the interface was fully removed) for a few milliseconds. UP notifications and reactions to removed interfaces (i.e. something that can be triggered by NB txn) is not delayed so that txns execute as fast as possible.

During the making of this PR I have also realized that contexts are ever more needed in descriptor methods (https://github.com/ligato/vpp-agent/issues/1569). With them (and some additional work) we will be able to get rid of or at least simplify the derived keys, which are getting really long and their parsing functions are complicated and error prone. For interface addresses and VRFs I had decided to carry all necessary information in the derived values instead of putting them into already complicated keys (e.g. https://github.com/milanlenco/vpp-agent/blob/master/plugins/linux/ifplugin/descriptor/interface.go#L664-L671). But this is only temporary and I plan to start on working on descriptor contexts and getting rid of all the complexity currently associated with derived values.

Btw. it was verified that PR doesn't break any e2e or integration tests for any of the supported VPP versions.

Signed-off-by: Milan Lenco <lencomilan@gmail.com>


tests/e2e/e2e.go
```diff
@@ -584,6 +587,37 @@ func (ctx *TestCtx) GetDerivedValueStateClb(baseValue proto.Message, derivedKey
...
+	}
+	q := fmt.Sprintf(`/scheduler/dump?key-prefix=%s&view=%v`, url.QueryEscape(model.KeyPrefix()), view)
+	resp, err := ctx.httpClient.GET(q)
+	if err != nil {
+		ctx.t.Fatalf("Request to dump values failed: %v", err)
+	}
+	var kmDump []KeyWithMetadata
+	if err := json.Unmarshal(resp, &kmDump); err != nil {
+		ctx.t.Fatalf("Reply with dumped values cannot be decoded: %v", err)
+	}
```
ondrej-fabry, line 612: Perhaps we could reuse the client for agentctl which provides this API for KVscheduler:

```suggestion
c, err := client.NewClient(client.DefaultAgentHost)
if err != nil {
	ctx.t.Fatal(err)
}
kmDump, err := c.SchedulerDump(context.Background(), types.SchedulerDumpOptions{
	KeyPrefix: model.KeyPrefix(),
	View:      view.String(),
})
```

and possibly further simplify it by instantiating the client during the Setup.

tests/e2e/080_vrf_test.go
```diff
@@ -455,3 +461,169 @@ func TestVRFRoutes(t *testing.T) {
...
+
+	// cleanup
+	req := ctx.GenericClient().ChangeRequest()
+	err = req.Delete(
+		existingVrf,
+		existingIface1,
+		iface2,
+	).Send(context.Background())
+	Expect(err).ToNot(HaveOccurred())
+	err = ifHandler.DeleteInterface(vrfIface1HostName)
```
ondrej-fabry, line 627: Why even store the error in a variable if it won't be checked? Better to either put `_ = ...` or check the error.

milanlenco: good catch, will add the forgotten check

milanlenco: fixed

tests/e2e/e2e.go
milanlenco: I have updated `e2e.go` to use client from agentctl for all interactions with the agent - please see the second commit.
I expect that there will be some merge conflicts with https://github.com/ligato/vpp-agent/pull/1770, but they should be relatively easy to fix.

ondrej-fabry (approved):

Issue #1767, opened 2020-12-04
Support for 3rd party configuration models in Agentctl


fgschwan:
## Problem
This issue is basically a more general and scope-wider version of [this](https://github.com/ligato/vpp-agent/issues/1740) issue.
For some agentctl commands to work properly, all the configuration models that the VPP-Agent is using must be known to agentctl. This is not the case for some agentctl commands. They depend on the [configurator.Config](https://github.com/ligato/vpp-agent/blob/effbcadbb8d02cf89caaf8bcbfd6d134eae9361d/proto/ligato/configurator/configurator.proto#L12-L16) that holds only the configuration models defined in the VPP-Agent repository. This is not a problem when no additional configuration models are needed. It is a problem when VPP-Agent is used as a base code in other code projects and more configuration models are added (and properly registered). These '3rd party' configuration models are missed out from consideration when some agentctl commands are executed. Here is the list of all agentctl commands depending on configurator.Config and need to support also 3rd party models:
 - `agentctl config delete`
 - `agentctl config retrieve`
 - `agentctl config watch`
 - `agentctl report` - list of affected report files inside report zip file:
    - `agent-NB-config.yaml`
    - `vpp-running-config(vpp-agent-SB-dump).yaml`

## Partial solution
The problem was already solved for `agentctl config get` and `agentctl config update` in [this](https://github.com/ligato/vpp-agent/pull/1754) PR (the same issue as mentioned above). Similar approach could be used for `agentctl config delete` (the [delete](https://github.com/ligato/vpp-agent/blob/effbcadbb8d02cf89caaf8bcbfd6d134eae9361d/proto/ligato/configurator/configurator.proto#L51-L63) in configurator.Config is basically the same as [update](https://github.com/ligato/vpp-agent/blob/effbcadbb8d02cf89caaf8bcbfd6d134eae9361d/proto/ligato/configurator/configurator.proto#L26-L45) so we can reuse update solution for delete easily) and `agentctl report`:`agent-NB-config.yaml` (this is just `agentctl config get` functionality).
The solution for other agentctl commands is not that as simple and is currently unknown.

chrismetz09:
Any anticipated changes to agentctl front-end? Or no change, performed under the covers?

fgschwan:
> 
> 
> Any anticipated changes to agentctl front-end? Or no change, performed under the covers?

[...]

PR #1769, opened 2020-12-08
feat: Added possibility to setup NB configuration from file at VPP-Agent start

fgschwan:
Added possibility to define NB configuration as file and to picked up and applied by VPP-Agent at start.
Additional features:
 - file defined as yaml
 - ability to set up also 3rd party models (using the same components as are used for `agentctl config update --replace`, see [this issue](https://github.com/ligato/vpp-agent/issues/1767) for understanding more about 3rd party model definition/problems)
 - by default active, no additional configuration needed to use (create `nb-initial-config.yaml` in vpp-agent working directory, i.e. if you installed vpp-agent binary in one of paths from env.variable then `nb-initial-config.yaml` must be in that directory where you are when you execute vpp-agent binary - if you are still not sure about the location, don't worry, VPP-agent will tell you in debug logging where it expects the file to be by using absolute file path)
 - fully configurable (path to configuration yaml file + feature on/off switch)
 - usable in projects using VPP-Agent as its base (using the updated code in VPP-Agent plugin setup will allow usage of this feature)
 - basic merging capabilities for NB configuration coming from different data sources (NB init file source is just another source among others like etcd and others) - in case of colliding data, other sources have priority to initial NB config file
 - e2e tests (start with NB init file, start with Etcd with data, start with both sources having data and merge colliding data - etcd data is used in the end)

### DIFF
--- a/plugins/orchestrator/localregistry/nbinitfileregistry.go
+++ b/plugins/orchestrator/localregistry/nbinitfileregistry.go
@@ -171,8 +171,12 @@ func (r *NBInitFileRegistry) retrieveConfig() (*Config, error) {
 	}
 	found, err := r.Cfg.LoadValue(config)
 	if !found {
-		r.Log.Debug("NBInitFileRegistry plugin config not found")
-		return config, nil
+		if err == nil {
+			r.Log.Debug("NBInitFileRegistry plugin config not found")
+		} else {
+			r.Log.Debugf("NBInitFileRegistry plugin config can't be loaded due to: %v", err)
+		}
+		return config, err
 	}
 	if err != nil {
 		return nil, err
--- a/plugins/orchestrator/localregistry/nbinitfileregistry.go
+++ b/plugins/orchestrator/localregistry/nbinitfileregistry.go
@@ -1,0 +1,286 @@
+// Copyright (c) 2020 Pantheon.tech
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at:
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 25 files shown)


fgschwan:
I ran e2e and integration test for VPP 20.01/20.05/20.09 and it passed for current PR state.

plugins/orchestrator/localregistry/nbinitfileregistry.go
```diff
@@ -0,0 +1,286 @@
...
+	"go.ligato.io/cn-infra/v2/db/keyval"
+	"go.ligato.io/cn-infra/v2/infra"
+	"go.ligato.io/vpp-agent/v3/client"
+	"google.golang.org/protobuf/encoding/protojson"
+	protoV2 "google.golang.org/protobuf/proto"
+)
+
+const (
+	registryName          = "nb-init-file-registry"
+	defaultNBInitFilePath = "nb-initial-config.yaml"
```
ondrej-fabry, line 39: I think we should lose the `nb-` prefix. There is no other initial config than NB - northbound one so does not add extra info for user, just makes the filename longer and can potentially confuse users.

fgschwan: Ok, that sound good. It also that means to change the name of the registry, its config name and content (i mean config for the registry not the init config). Otherwise, that will confuse anyone looking into the code.

ondrej-fabry: Yes, alright

fgschwan: Fixed in [`d7d34d3` (#1769)](https://github.com/ligato/vpp-agent/pull/1769/commits/d7d34d35dce453592232e386f79f677bdd5bfd86)

fgschwan:
I ran e2e and integration test for VPP 20.01/20.05/20.09 and it passed for current PR state.

fgschwan:
Rebased changes from merged e2e test changes (https://github.com/ligato/vpp-agent/pull/1770).
Unit/integration/e2e tests for vpp 20.05 passed ok.

ondrej-fabry (approved):

milanlenco (approved):

PR #1779, opened 2021-01-19
feat: Support for cache DNS server in VPP + e2e test refactoring

fgschwan:
I will add only the simple functionality (turn it on and configure the upstream DNS server that will VPP consult for unknown domain names).
However, this PR will contain also e2e test and related e2e changes. These related changes touched some basic things that lead to some heavy e2e refactoring (docker contrainer code deduplication, decupling of basic test topology components (agent,etcd,microservices,...) with underlaying runtime (docker containers)).


fgschwan:
Unit test + integration tests (vpp 20.01,20.05,20.09) + e2e tests (vpp 20.01,20.05,20.09) were successful.

tests/e2e/options.go
```diff
@@ -107,26 +199,54 @@ func DefaultAgentOpt() *AgentOpt {
...
 	return func(o *SetupOpt) {
 		o.SetupAgent = false
 	}
 }
 
-// WithEtcd is test setup option enabling vpp-agent setup
+// WithCustomVPPAgent is test setup option using alternative vpp-agent image (customized original vpp-agent)
+func WithCustomVPPAgent() SetupOptModifier {
+	return func(o *SetupOpt) {
+		o.AgentOpt.Image = "vppagent.test.ligato.io:custom"
```
ondrej-fabry, line 220: Shouldn't this option accept parameter to allow defining truly _custom_ image?

fgschwan: I'm using the [customized vpp-agent image made in e2e script](https://github.com/ligato/vpp-agent/pull/1779/files#diff-f55dfa5c52801afc2b7175e0e2343b181f02bd2329837e80e769f2498c213cd7R44-R48). This allows me to prepare this image only once per run of all e2e tests. If i make this method truly custom by introducing options, then i will have to prepare the image in test runtime. That means image creation per test and not per one test suite run. If 2 test want to use the custom image(doesn't happen now, but could in future), they do not know whether it was rebuilded using the latest vpp-agent image as base image or not. So it gets rebuilded multiple times. So i choose the easy way to rebuild it only once before test suite run, but the downside is that it is hardcoded.

plugins/vpp/dnsplugin/descriptor/dnsserver.go
```diff
@@ -0,0 +1,149 @@
...
+				"address %s, should be a valid ipv4/ipv6 address", serverIpAddress), "upstreamDnsServers")
+		}
+	}
+	return nil
+}
+
+// EquivalentDNSServers determines whether 2 DNS servers are logically equal. This comparison takes
+// into consideration also semantics that couldn't be modeled into proto models (i.e. upstream servers are
+// IP addresses and not only strings)
+func (d *DNSServerDescriptor) EquivalentDNSServers(key string, oldDNSServer, newDNSServer *dns.DNSServer) bool {
```
milanlenco, line 80: EquivalentDNSServers could compare UpstreamDnsServers as sets, i.e. different order of the same set of values could be considered equivalent. Otherwise it behaves just like proto.Equal and so it is not neccessarry.

plugins/vpp/dnsplugin/descriptor/dnsserver.go
```diff
@@ -0,0 +1,149 @@
...
+// Update updates DNS Server configuration of already enabled DNS functionality
+func (d *DNSServerDescriptor) Update(key string, oldValue, newValue *dns.DNSServer,
+	oldMetadata interface{}) (newMetadata interface{}, err error) {
+	if err := d.updateUpstreamDNSServerList(oldValue.UpstreamDnsServers, newValue.UpstreamDnsServers); err != nil {
+		return nil, errors.Errorf("can't update upstream DNS servers due to: %v", err)
+	}
+	return nil, nil
+}
+
+func (d *DNSServerDescriptor) updateUpstreamDNSServerList(oldServers, newServers []string) error {
```
milanlenco, line 127: Could be more efficient - delete only the obsolete entries and add only the truly new ones. Otherwise it is essentially re-create and so the Update method could be omitted.

proto/ligato/vpp/dns/models.go
```diff
@@ -0,0 +1,30 @@
...
+
+const (
+	ModuleName = "vpp.dns"
+)
+
+var (
+	// DNSServer is registered NB model of DNSServer
+	ModelDNSServer = models.Register(&DNSServer{}, models.Spec{
+		Module:  ModuleName,
+		Type:    "dnsserver",
```
milanlenco, line 27: I would maybe go with type "dnscache" (likewise for proto message name), just to make it clear that this is not actually a proper DNS server.
Version can be "v1" since this is the first version of the model.

fgschwan: Fixed in [`34ea247`](https://github.com/ligato/vpp-agent/pull/1779/commits/34ea247795b517addafa51056f45b2a4a6d5cba2)

plugins/vpp/dnsplugin/descriptor/dnsserver.go
fgschwan: Upstream DNS server is a list and not a set. Removing EquivalentDNSServers. Fixed in [`ac77bf2`](https://github.com/ligato/vpp-agent/pull/1779/commits/ac77bf2faa7c4b796216c7917a545284c73d97d3)

plugins/vpp/dnsplugin/descriptor/dnsserver.go
fgschwan: Initially i wanted to do it efficiently until i realized that it is not possible. VPP doesn't allow me to insert new upstream DNS server into proper place in the list, but just append at the end of list. Hence the recreate. However i forgot to reevaluate the usefulness of the Update function implementation.
Fixed in [`ac77bf2`](https://github.com/ligato/vpp-agent/pull/1779/commits/ac77bf2faa7c4b796216c7917a545284c73d97d3).

milanlenco (approved):

ondrej-fabry (approved):

PR #1782, opened 2021-01-28
fix: fixed delete by resync failure of init file data by using agentctl

fgschwan:
When using init file for importing configuration data by VPP-Agent start and then you try to delete them by resync with empty configuration (using agentctl) then failure happens (this is not the case for using agentctl config delete). The root cause of this is different data source of init file data and data that can agentctl work with. The data source behaves for resyncs similar to linux namespaces, you can't clean data from different data source/namespace (agentctl works with 'grpc' data source data and init file data are 'datasync'-data-source data).

The fix provided by this PR enables the user to configure the init file data to have different data source (i.e.'grpc' to be compatible with agentctl resync). However, due to aggregating of data from multiple data sources (aggregator functionality), the data source will be changes also for data from other sources (etcd, redis,...). That means that this fix is only for certain scenarios when it is used only init file and no etcd, redis,... . 

Proper handling of all possible combinations (init file/init file + etcd/init file + redis/...) to achieve functionality of agentctl's resync ('agentctl update --replace') is out of scope of this fix. It would probably require to do multiple resync at VPP-Agent start (multiple data inputs can have different data source labels -> for each one must be done resync) and handle all corner cases.

### DIFF
--- a/plugins/orchestrator/watcher/aggregator.go
+++ b/plugins/orchestrator/watcher/aggregator.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"strings"
 
+	"go.ligato.io/cn-infra/v2/config"
 	"go.ligato.io/cn-infra/v2/datasync"
 	"go.ligato.io/cn-infra/v2/datasync/kvdbsync"
 	"go.ligato.io/cn-infra/v2/datasync/kvdbsync/local"
@@ -27,6 +28,7 @@ import (
 	"go.ligato.io/cn-infra/v2/infra"
 	"go.ligato.io/cn-infra/v2/logging"
 	"go.ligato.io/cn-infra/v2/utils/safeclose"
+	"go.ligato.io/vpp-agent/v3/plugins/orchestrator/contextdecorator"
 	"go.ligato.io/vpp-agent/v3/plugins/orchestrator/localregistry"
 )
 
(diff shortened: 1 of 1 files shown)


milanlenco:
Would it be possible to disassociate init-file from datasync (kvdb), i.e. to be able to have different data source for init file than to have for e.g. etcd?
It seems that it is possible to put data-source into the context of change/resync event:
https://github.com/ligato/vpp-agent/blob/master/plugins/orchestrator/orchestrator.go#L255-L262
https://github.com/ligato/vpp-agent/blob/master/plugins/orchestrator/orchestrator.go#L209-L216

If it would be possible to set context from initfileregistry.go then the option to customize data source of init-file could go to `initfileregistry.conf`.

fgschwan:
> 
> 
> Would it be possible to disassociate init-file from datasync (kvdb), i.e. to be able to have different data source for init file than to have for e.g. etcd?
> It seems that it is possible to put data-source into the context of change/resync event:
> [https://github.com/ligato/vpp-agent/blob/master/plugins/orchestrator/orchestrator.go#L255-L262](https://github.com/ligato/vpp-agent/blob/master/plugins/orchestrator/orchestrator.go?rgh-link-date=2021-01-28T11%3A31%3A24Z#L255-L262)
> [https://github.com/ligato/vpp-agent/blob/master/plugins/orchestrator/orchestrator.go#L209-L216](https://github.com/ligato/vpp-agent/blob/master/plugins/orchestrator/orchestrator.go?rgh-link-date=2021-01-28T11%3A31%3A24Z#L209-L216)
> 
> If it would be possible to set context from `initfileregistry.go` then the option to customize data source of init-file could go to `initfileregistry.conf`.

I use local client to set the init file data into local registry. If its API had the ability to set the transaction commit context(it has NOT right now), then i could decorate it with datasource and when it comes to aggregator (i think it should pass the context as is, but didn't checked in runtime, just some code checking) then i could do some different actions for data from initfile. However, this can be done also by some hardwiring in aggregator (local client data are already hardwired with different logic). This of course doesn't solve the out of scope issue to do multiple resyncs due to multiple data sources. 
If the purpose is just to move the configuration to initfile config, then this would be misleading, because the context labeling with data source happens for all data inputs of the agregator at level where are the changes merged. So it doesn't effect only the initfile data, but all data. Hence, the new config.

milanlenco (approved):

PR #1784, opened 2021-02-05
feat: Added REST API for retrieving JSON schema for VPP-Agent configuration

fgschwan:
I also included 3rd party proto-to-jsonschema converter into the code due to problematic usage as external library and possible future changes/customizations of schema.


plugins/restapi/handlers.go
```diff
@@ -315,6 +326,123 @@ func (p *Plugin) registerHTTPHandler(key, method string, f func() (interface{},
...
 }
 
+// jsonSchemaHandler returns JSON schema of VPP-Agent configuration.
+func (p *Plugin) jsonSchemaHandler(formatter *render.Render) http.HandlerFunc {
+	return func(w http.ResponseWriter, req *http.Request) {
+		// create FileDescriptorProto for dynamic Config holding all VPP-Agent configuration
+		knownModels, err := client.LocalClient.KnownModels("config") // locally registered models
+		if err != nil {
+			errMsg := fmt.Sprintf("500 Internal server error: "+
+				"can't get registered models: %v\n", err)
```
rewenset, line 336: looks a bit strange to duplicate HTTP status code and its text in the response body. I think, you may omit that `"500 Internal server error: "` or is it required? :thinking:

plugins/restapi/handlers.go
```diff
@@ -315,6 +326,123 @@ func (p *Plugin) registerHTTPHandler(key, method string, f func() (interface{},
...
+			",proto_and_json_fieldnames" // allow also proto names for fields
+		cgReq := &protoc_plugin.CodeGeneratorRequest{
+			ProtoFile:       fileDescriptorProtos,
+			FileToGenerate:  []string{dynConfigFileDescProto.GetName()},
+			Parameter:       &params,
+			CompilerVersion: nil, // compiler version is not need in this protoc plugin
+		}
+		cgReqMarshalled, err := proto.Marshal(cgReq)
+		if err != nil {
+			fmt.Errorf("can't proto marshal to: %w", err)
```
rewenset, line 368: probably a typo that return value is not used

fgschwan: It is forgotten leftover from error handling changes. Fixed in [`cd09327`](https://github.com/ligato/vpp-agent/pull/1784/commits/cd093279785a4b0ce46f06cc6b03e9d21bd12a76).

plugins/restapi/handlers.go
fgschwan: It was copy/paste of error handling from other rest handlers, but i can omit that. It looks strange and the user of my REST API addition don't need that. However, don't know about other handlers. I guess it is only good for logging and not for response (logging and response body use the same variable). So i preserved it for logging and left out from http response message body.
Fixed in [`cd09327`](https://github.com/ligato/vpp-agent/pull/1784/commits/cd093279785a4b0ce46f06cc6b03e9d21bd12a76).

chrismetz09:
Will this REST API return the same information as agentctl config get or retrieve?

fgschwan:
> 
> 
> Will this REST API return the same information as agentctl config get or retrieve?

@chrismetz09 No. This PR adds JSON schema retrieval. The JSON schema only describes JSON (or YAML converted to JSON). It doesn't get actual configuration data (as returned by agentctl config get), but it does describes the structure returned by agentctl config get (when retrieved yaml is converted to json). The reason for this REST API is the ability to verify validity of any given yaml configuration for VPP-Agent (yaml is converted to JSON and JSON is validated against JSON schema).

milanlenco (approved):
Tested also with StoneWork (private project where some models are proxied) and it works without issues.
In the future we could maybe also expose this via agentctl (i.e. add command to obtain the schema).

One unfortunate thing (apart from missing descriptions) is that multi-word fields are twice in the schema to allow both snake and camel case styles (because both would be accepted by the proto/json unmarshaler):
```
                            "vpp_interface": {
                                "type": "string"
                            },
                            "vppInterface": {
                                "type": "string"
                            },
```
I will look into how we can avoid this. So for found this: https://schema.org/alternateName
But it is probably not what I would hope for (tested with online validators).
Also I might look into how we can get the descriptions because it would be helpful for the documentation purposes (which I'm currently invested in).

fgschwan:
> One unfortunate thing (apart from missing descriptions) is that multi-word fields are twice in the schema to allow both snake and camel case styles (because both would be accepted by the proto/json unmarshaler):
> 
> ```
>                             "vpp_interface": {
>                                 "type": "string"
>                             },
>                             "vppInterface": {
>                                 "type": "string"
>                             },
> ```
@milanlenco This is configurable by the 3rd party tool that i included ([tool code](https://github.com/ligato/vpp-agent/pull/1784/files#diff-62efd319b445478517819dc7e3b9f1f0aff86e1187a103c643110b619fd0c230R484-R493)(expand the file plugins/restapi/jsonschema/converter/types.go to see it)), but i hardcoded the [tool usage code](https://github.com/ligato/vpp-agent/pull/1784/files#diff-ed5c24ad22126925d11387ccf4f7c278b09fd512aa6376ef5099ce00bb93e311R359) to show both. Yaml and related JSON used in agentctl config get/update can use both so the schema should have both. 
However i can understand that this might introduce some unexpected problems with other tools that use the JSON schema and reasonable resolution could be to limit usage(from 3rd party projects view using VPP-Agent as library) just to snail case or the other format and get only json schema with one of these formats. So do you prefer that the type of field name usage should be configurable (VPP-Agent configuration) and by default it should behave as it is right now (show both)?

milanlenco:
> > One unfortunate thing (apart from missing descriptions) is that multi-word fields are twice in the schema to allow both snake and camel case styles (because both would be accepted by the proto/json unmarshaler):
> > ```
> >                             "vpp_interface": {
> >                                 "type": "string"
> >                             },
> >                             "vppInterface": {
> >                                 "type": "string"
> >                             },
> > ```
> 
> @milanlenco This is configurable by the 3rd party tool that i included ([tool code](https://github.com/ligato/vpp-agent/pull/1784/files#diff-62efd319b445478517819dc7e3b9f1f0aff86e1187a103c643110b619fd0c230R484-R493)(expand the file plugins/restapi/jsonschema/converter/types.go to see it)), but i hardcoded the [tool usage code](https://github.com/ligato/vpp-agent/pull/1784/files#diff-ed5c24ad22126925d11387ccf4f7c278b09fd512aa6376ef5099ce00bb93e311R359) to show both. Yaml and related JSON used in agentctl config get/update can use both so the schema should have both.
> However i can understand that this might introduce some unexpected problems with other tools that use the JSON schema and reasonable resolution could be to limit usage(from 3rd party projects view using VPP-Agent as library) just to snail case or the other format and get only json schema with one of these formats. So do you prefer that the type of field name usage should be configurable (VPP-Agent configuration) and by default it should behave as it is right now (show both)?

Maybe the REST API for json schema could have an option to print the schema for snake case or camel case or both (default).
For example in documentation I would present just one of these styles (as the preferred one) and mention that the other one is also accepted.

fgschwan:
> Maybe the REST API for json schema could have an option to print the schema for snake case or camel case or both (default).
> For example in documentation I would present just one of these styles (as the preferred one) and mention that the other one is also accepted.

Implemented as URL Query paramter in [`2a816d4`](https://github.com/ligato/vpp-agent/pull/1784/commits/2a816d477f1d6ad57ca21b9e7d631e8ae43c5a8f).
Use these URLs to change the field name generation in JSON schema export:
```
<VPP-Agent IP address>:9191/info/configuration/jsonschema
<VPP-Agent IP address>:9191/info/configuration/jsonschema?fieldnames=onlyjson
<VPP-Agent IP address>:9191/info/configuration/jsonschema?fieldnames=onlyproto
```

milanlenco (approved):

PR #1785, opened 2021-02-09
feat: Add support for VPP 21.01

milanlenco:
This PR adds support for the latest VPP release 21.01

The major change is the NAT44 plugin refactor. For example it is now required to explicitly enable and provide initial configuration to the NAT44 plugin via binary API before it can be used. Previously the initial configuration was given inside the VPP startup configuration (`/etc/vpp/vpp.conf`) and couldn't be changed in the run-time. Also the plugin was always enabled and apparently was taking up some resources.

For VPP 21.01 and newer the NAT44 plugin will be enabled by submitting the `NAT44Global` configuration item (which had to be configured anyway to enable forwarding). Also [endpoint dependency is now possible to configure and change via NB API](https://github.com/milanlenco/vpp-agent/blob/vpp2101/proto/ligato/vpp/nat/nat.proto#L16-L29). I decided to select the endpoint-dependent mode as the default (that was also the case before with our `/etc/vpp/vpp.conf`). The model has "endpoint-**in**dependent" field (with `false` as default) as opposed to endpoint-dependent because in vast majority of cases endpoint dependency is preferred and most features of NAT44 do not work in independent mode anyway (so to avoid complains empty global NAT44 config will enable all features that can be configured via ligato). Aside from endpoint dependency there are few more init options for NAT44 and we could add them later into the model as needed. 

Also output from "show memory" was slightly changed and the RE that we use to parse it had to be modified for 21.01.

Aside for new features/changes there are also new bugs as always:
- [](https://jira.fd.io/browse/VPP-1968) - (a fix is already prepared and ready for review: https://gerrit.fd.io/r/c/vpp/+/31172)
- [](https://jira.fd.io/browse/VPP-1969)
- [](https://jira.fd.io/browse/VPP-1970)

Support for VPP 19.08 was removed.

All integration and e2e tests are passing for all four supported VPP versions. 
:warning: However, please note that some checks are disabled (skipped) for VPP 21.01 until the bugs mentioned above are fixed and merged.


chrismetz09:
Docs Update: add `endpoint_independent` to NAT Global REST API at https://docs.ligato.io/en/latest/api/api-vpp-agent/#vpp-nat-global?

milanlenco:
> Docs Update: add `endpoint_independent` to NAT Global REST API at https://docs.ligato.io/en/latest/api/api-vpp-agent/#vpp-nat-global?

The example output in docs can remain as is. `endpoint_independent` will be in the vast majority of cases `false` and therefore not visible in the output from REST API.

examples/tutorials/07_vpp-connection/main.go
```diff
@@ -152,10 +152,11 @@ func (p *HelloWorld) Close() error {
 	return nil
 }
 
-func macParser(mac string) []byte {
+func parseMac(mac string) (parsedMac [6]uint8) {
```
ondrej-fabry, line 155: This should probably be replaced with generated helper functions for parsing MAC strings.

plugins/vpp/ifplugin/vppcalls/vpp2101/bond_vppcalls.go
```diff
@@ -0,0 +1,118 @@
+//  Copyright (c) 2019 Cisco and/or its affiliates.
```
rewenset, line 1: What is a guideline for a copyright year? Should it be set to the current year only for new files or for updated files too?

examples/tutorials/07_vpp-connection/main.go
milanlenco: Fixed in https://github.com/ligato/vpp-agent/pull/1785/commits/533ad8fb93433cfe8a9d6d482365be935e58fea9

plugins/vpp/ifplugin/vppcalls/vpp2101/bond_vppcalls.go
milanlenco: If a file is only copied/renamed, like in this case, then I would leave the copyright header as is.

PR #1792, opened 2021-03-26
fix: VPP interface requires IP address with network prefix

milanlenco:


juliusmilan (approved):
+1

fgschwan (approved):

PR #1858, opened 2022-07-13
Add support for agentctl configuration labels

pemoticak:
This PR adds support for user defined tags that can be associated with configuration labels.

The labels can be specified using the `--labels` flag with `agentctl config update`.
Similarly, the output of `agentctl config get` can be filtered by specifying the `--labels` flag (it will show only those config items that have the given tags associated with them).
The tags are currently specified as a comma separated list of key=value pairs e.g. `--labels="foo=bar","baz=qux"`.

Signed-off-by: Peter Motičák <peter.moticak@pantheon.tech>


client/client_api.go
```diff
@@ -48,6 +62,13 @@ type GenericClient interface {
 	// TODO: return as list of config items
 	GetConfig(dsts ...interface{}) error
 
+	// GetFilteredConfig retrieves current config into dsts according to the provided filter.
+	GetFilteredConfig(filter Filter, dsts ...interface{}) error
+
+	UpdateConfig(ctx context.Context, items UpdateItems) (*generic.SetConfigResponse, error)
+
+	DeleteConfig(ctx context.Context, items UpdateItems) (*generic.SetConfigResponse, error)
```
ondrej-fabry, line 70: These are not much of an user-friendly improvement for the original gRPC service definition.. it's practically almost same if the user uses this methods or the generated gRPC [`generic.ManagerServiceClient`](https://github.com/ligato/vpp-agent/blob/master/proto/ligato/generic/manager_grpc.pb.go#L24). And it's actually worse for user to use this over the generated client..

My top concerns about this are:
- field `OverwriteAll` in the `UpdateItems` struct
  - introduces another name for the same operation: Resync/Replace/OverwriteAll (it is unfortuante that different parts of code use different name for this, but let's stick with single one per entity)
  - this option does not make sense in `DeleteConfig` method (replace and delete are mutually exclusive operations), instead it should be a parameter in the `UpdateConfig` method or simply as another method.. 
- type `generic.SetConfigResponse` used as return value in the methods of client interface
  - this forces non-gRPC implementation (localclient) to return this gRPC response wrapper type (currently returns `nil` though)
  - this should return slice of update results
- `UpdateItems` contains list of proto messages (items) and labels
  - this makes it impossible to apply labels per item in a single update even though the actual gRPC API allows this (rpc `SetConfig` containss list of `UpdateItem`s - _item + labels_ 
- the client interface is still missing universal method for getting entire config as return value (list of items)
- I think the `Filter` should only accept a single way to filter especially when filtering by IDs
  - user either wants to get specific items by referring to their ID 
  - or he wants to get all items that match the requested labels
  - allowing both filters in one request might be confusing about what we expect server to return

plugins/orchestrator/meta_service.go
```diff
@@ -35,6 +35,13 @@ import (
 	"go.ligato.io/vpp-agent/v3/proto/ligato/generic"
 )
 
+var reservedLabelKeys = map[string]struct{}{
```
ondrej-fabry, line 38: I think we can omit enforcing of the reserved label keys. Let's just define some guidelines for the key format. I would simply use same guidelines as the [Docker Labels](https://docs.docker.com/config/labels-custom-metadata/#key-format-recommendations):

> ### Key format recommendations
> 
> A label key is the left-hand side of the key-value pair. Keys are alphanumeric strings which may contain periods (`.`) and hyphens (`-`). Most Docker users use images created by other organizations, and the following guidelines help to prevent inadvertent duplication of labels across objects, especially if you plan to use labels as a mechanism for automation.
> 
> - Authors of third-party tools should prefix each label key with the reverse DNS notation of a domain they own, such as `com.example.some-label`.
> 
> - Do not use a domain in your label key without the domain owner’s permission.
> 
> - The `com.docker.*`, `io.docker.*`, and `org.dockerproject.*` namespaces are reserved by Docker for internal use.
> 
> - Label keys should begin and end with a lower-case letter and should only contain lower-case alphanumeric characters, the period character (`.`), and the hyphen character (`-`). Consecutive periods or hyphens are not allowed.
> 
> - The period character (`.`) separates namespace “fields”. Label keys without namespaces are reserved for CLI use, allowing users of the CLI to interactively label Docker objects using shorter typing-friendly strings.
> 
> These guidelines are not currently enforced and additional guidelines may apply to specific use cases.

Where the keys for internal use would be `io.ligato.*` and `ligato.*`

chrismetz09:
@pemoticak @ondrej-fabry 
 In prep to update [agentctl docs](https://docs.ligato.io/en/latest/user-guide/agentctl/), need clarification and example of labels in action. I looked over `100_agentctl_test.go` and see this test config:

test config:
```
`config/vpp/v2/interfaces/tap1 {"name":"tap1", "type":"TAP", "enabled":true, "ip_addresses":["10.10.10.10/24"], "tap":{"version": "2"}}`,
```
Then I see some labels:
```
...
updateLabels := []string{"if=dummy", "\"if=dummy\",\"source=test\"", "\"if=differentvalue\",\"source=test\"", "", "\"onlykey=\""}
```
Are all of these labels associated with the the ../interfaces/tap1 config item? Stored where? Contained in a `.yaml` config file? I will scan code after PR merge and run `agentctl ... --labels=...` to test it out and hack docs for it. More questions might follow.

cmd/agentctl/commands/config.go
```diff
@@ -71,11 +72,13 @@ func newConfigGetCommand(cli agentcli.Cli) *cobra.Command {
 	}
 	flags := cmd.Flags()
 	flags.StringVarP(&opts.Format, "format", "f", "", "Format output")
+	flags.StringSliceVar(&opts.Labels, "labels", []string{}, "Output only config items that have the given labels")
```
ondrej-fabry, line 75: Describe the accepted format for the label key/value pairs.

cmd/agentctl/commands/config.go
```diff
@@ -135,6 +138,7 @@ func newConfigUpdateCommand(cli agentcli.Cli) *cobra.Command {
 	// flags.BoolVarP(&opts.Verbose, "verbose", "v", false, "Show verbose output")
 	flags.DurationVarP(&opts.Timeout, "timeout", "t",
 		5*time.Minute, "Timeout for sending updated data")
+	flags.StringSliceVar(&opts.Labels, "labels", []string{}, "Labels associated with updated config items")
```
ondrej-fabry, line 141: Same here.

plugins/orchestrator/store.go
```diff
@@ -29,14 +29,30 @@ type KVStore interface {
...
+// KLStore describes an interface for key-label store used by dispatcher.
+type KLStore interface {
+	ListLabels(key string) Labels
+	AddLabel(key, lkey, lval string)
+	HasLabel(key, lkey string) bool
+	DeleteLabel(key, lkey string)
+	ResetLabels(key string)
+}
+
+type KStore interface {
```
ondrej-fabry, line 41: I think keeping the `K` as suffix here does not make much sense anymore. To avoid confusion might be better to just name it `Store`.

chrismetz09: agree

ondrej-fabry (approved):

PR #1860, opened 2022-08-03
fix: agentctl config history not working with remote models

pemoticak:
When calling agentctl config history, just query the remote models with
grpc client and add them to the global default model registry.
Then when we unmarshal the messages we primarily use the default registry and
the proto global registry is used as a fallback.

Signed-off-by: Peter Motičák <peter.moticak@pantheon.tech>

### DIFF
--- a/plugins/kvscheduler/internal/utils/record.go
+++ b/plugins/kvscheduler/internal/utils/record.go
@@ -22,6 +22,8 @@ import (
 	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/reflect/protoreflect"
 	"google.golang.org/protobuf/reflect/protoregistry"
+
+	"go.ligato.io/vpp-agent/v3/pkg/models"
 )
 
 // RecordedProtoMessage is a proto.Message suitable for recording and access via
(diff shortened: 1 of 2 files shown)


plugins/kvscheduler/internal/utils/record.go
```diff
@@ -77,10 +79,17 @@ func (p *RecordedProtoMessage) UnmarshalJSON(data []byte) error {
 	if p.ProtoMsgName == "" {
 		return nil
 	}
-	msgType, err := protoregistry.GlobalTypes.FindMessageByName(protoreflect.FullName(pwn.ProtoMsgName))
+
+	typeRegistry := models.DefaultRegistry.MessageTypeRegistry()
+	fullMsgName := protoreflect.FullName(pwn.ProtoMsgName)
+	msgType, err := typeRegistry.FindMessageByName(fullMsgName)
+	if err != nil {
+		msgType, err = protoregistry.GlobalTypes.FindMessageByName(fullMsgName)
```
ondrej-fabry, line 87: Please add a comment here that this lookup in global proto registry is used as a fallback if proto model is not found in models registry.

ondrej-fabry (approved):

PR #1863, opened 2022-08-09
fix: agentctl config dump not working with remote models

pemoticak:
#### This PR:
- fixes `agentctl config dump` to work with remote models (fix is very similar to #1860)
- refactors the scheduler client API to use a type more suited for JSON (un)marshalling
- recorded proto messages are now (un)marshalled using protojson package instead of prototext
- ModelList now registers the models that it fetches into the DefaultRegistry

These changes (mainly the ModelList one) fix some problems with trailing slashes in remote model key prefixes (for example when running `agentctl model list/inspect`).


ondrej-fabry:
is this ready for merge?

pemoticak:
Yes the fix itself works and should be ready for merge but it doesn't have an e2e test. The test could be added in a separate PR at a later time though.

plugins/kvscheduler/api/txn_record.go
```diff
@@ -172,6 +172,11 @@ type RecordedTxnOps []*RecordedTxnOp
 // RecordedTxns is a list of recorded transactions.
 type RecordedTxns []*RecordedTxn
 
+type RecordedKVWithMetadata struct {
```
fgschwan, line 175: Just add some explaining godoc for this new structure. When i dig deeper for the difference between this and KVWithMetadata, i got only added ProtoMsgName field (in RecordedProtoMessage struct somewhere in internal). So please state clearly why we need it, what is the added value (+maybe compare with KVWithMetadata).

fgschwan (changes requested):

fgschwan (approved):

PR #1870, opened 2022-09-14
feat: Multi agent tests

pemoticak:
This PR allows every instance of vpp-agent to have its own client and therefore it can be controlled/configured (before only a single default agent had its own client). It also changes how the e2e agents/microservices are started/stopped.

This prepares the e2e framework for future test cases and for the separation of framework into separate package (away from e2e tests themselves).


tests/e2e/vppagent.go
```diff
@@ -61,9 +83,33 @@ func startAgent(ctx *TestCtx, name string, opts *AgentOpt) (*Agent, error) {
 	if err != nil {
 		return nil, errors.Errorf("can't start agent %s due to: %v", name, err)
 	}
+	client, err := ctl.NewClient(agent.IPAddress())
```
fgschwan, line 86: No need for temporal `client` variable, just `agent.Client,err = ctl.NewClient(...)`

tests/e2e/vppagent.go
```diff
@@ -15,36 +15,58 @@
...
 )
 
 var vppPingRegexp = regexp.MustCompile("Statistics: ([0-9]+) sent, ([0-9]+) received, ([0-9]+)% packet loss")
 
 // Agent represents running VPP-Agent test component
 type Agent struct {
 	ComponentRuntime
-	ctx  *TestCtx
-	name string
+	Client ctl.APIClient
```
fgschwan, line 57: Exported field again. Is it needed? Definitely not now and i don't see the future changes as there is no additional PR for them. So no visible need to make it exported in this PR.

fgschwan (changes requested):
2 of 4 comments are related to variable exposure that might be valid in context of future changes (unknown to me) but could be also mistakes/not used in future changes so it is hard to accept them for current state of code.

tests/e2e/e2e.go
```diff
@@ -45,24 +41,17 @@ import (
...
 const (
-	agentInitTimeout     = time.Second * 15
-	processExitTimeout   = time.Second * 3
 	checkPollingInterval = time.Millisecond * 100
 	checkTimeout         = time.Second * 6
-	defaultTestShareDir  = "/test-share"
+	shareDir             = "/test-share"
 	shareVolumeName      = "share-for-vpp-agent-e2e-tests"
-	nameOfDefaultAgent   = "agent0"
+	MainAgentName        = "agent0"
```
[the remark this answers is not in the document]
pemoticak: > Maybe in future we need to expose it when package structure change and all, but not now.
Yes the original plan was to separate the e2e test harness from the tests themselves so they are not in the same package.
But that will be done in a separate PR, so I agree exporting these constants (and the client field below) in this PR is a mistake and I am reverting the export changes.

tests/e2e/etcd.go
```diff
@@ -58,10 +58,6 @@ func NewEtcd(ctx *TestCtx, optMods ...EtcdOptModifier) (*Etcd, error) {
 	return etcd, nil
 }
 
-func (ec *Etcd) Stop(options ...interface{}) error {
-	return ec.ComponentRuntime.Stop(nil, options)
```
fgschwan: Etcd container still need to be stopped and removed, right? So the ComponentRuntime.Stop is here still needed here.

fgschwan (changes requested):

tests/e2e/etcd.go
pemoticak: There is no context cleanup needed for the etcd container, so the ComponentRuntime.Stop function can be used directly via interface embedding.

P.S. The e2e tests were failing because I forgot to change the cleanup of the microservice component (this should be fixed in f69d594).

fgschwan (approved):

tests/e2e/e2e.go
```diff
@@ -363,143 +316,101 @@ func (test *TestCtx) VppRelease() string {
 }
 
 // GenericClient provides generic client for communication with default VPP-Agent test component
+// Deprecated: use ctx.Agent.GenericClient(...) instead
```
ondrej-fabry, line 319: I would not mark this as deprecated. Since most of the test cases use a single agent, using this shorter call for the main agent makes it much more easier to read keeps.

tests/e2e/010_interfaces_test.go
```diff
@@ -98,13 +98,13 @@ func TestInterfaceConnTap(t *testing.T) {
 
 	// restart microservice twice
 	for i := 0; i < 2; i++ {
-		ctx.StopMicroservice(msName)
+		ctx.Expect(ms.Stop()).To(Succeed())
```
ondrej-fabry, line 101: Could this check happen inside the `Stop()` function. This call is operating on the test setup and is not really part of this test case - just part of the test framework needed to run the test so it makes more sense to make the error checking in the background.

pemoticak: Agreed, checking the error inside the test itself isn't very elegant. The problem is the same function is used inside the test setup/harness itself, where returning error and then checking it is potentially useful (agent.Stop() works the same and can return different errors). After test cases and test harness are separated into different packages, the solution could be to have unexported  stop() function that returns an error (for use inside the test harness package) and then one exported function that doesn't return an error (for use in test cases). 
For now in the test cases we could: 

- revert to using ctx.StopMicroservice 
- not check the error at all 
- leave it as is

The first solution seems the best to me until the split happens in separate PR.

tests/e2e/e2e.go
```diff
@@ -363,143 +316,101 @@ func (test *TestCtx) VppRelease() string {
...
-	if err != nil {
-		test.t.Fatalf("Failed to get generic VPP-agent client: %v", err)
-	}
-	return c
+	test.t.Helper()
+	return test.Agent.GenericClient()
 }
 
 // GRPCConn provides GRPC client connection for communication with default VPP-Agent test component
+// Deprecated: use ctx.Agent.GRPCConn(...) instead
```
ondrej-fabry, line 325: The previous comment applies to this comment as well.. and the others that basically provide quick access to methods of the main agent.

tests/e2e/e2e.go
```diff
@@ -363,143 +316,101 @@ func (test *TestCtx) VppRelease() string {
...
-	}
-	return txn.Executed
+	test.t.Helper()
+	return test.Agent.GRPCConn()
 }
 
-// AgentInSync checks if the agent NB config and the SB state (VPP+Linux)
-// are in-sync.
+// AgentInSync checks if the agent NB config and the SB state (VPP+Linux) are in-sync.
+// Deprecated: use ctx.Agent.IsInSync(...) instead
```
ondrej-fabry, line 332: ^

tests/e2e/e2e.go
```diff
@@ -363,143 +316,101 @@ func (test *TestCtx) VppRelease() string {
...
+// StopAgent stops VPP-Agent with given name
+// Deprecated: use agent := ctx.StartAgent(name, opts...); agent.Stop()
 func (test *TestCtx) StopAgent(name string) {
 	test.t.Helper()
 
 	agent, found := test.agents[name]
 	if !found {
 		// bug inside a test
-		test.t.Logf("ERROR: cannot stop unknown agent %q", name)
+		test.t.Logf("ERROR: cannot stop unknown microservice %s", name)
```
ondrej-fabry, line 409: This was changed by accident ?

tests/e2e/e2e.go
```diff
@@ -517,6 +428,15 @@ func (test *TestCtx) AlreadyRunningMicroservice(msName string) *Microservice {
 	return ms
 }
 
+func (test *TestCtx) AlreadyRunningAgent(agentName string) *Agent {
```
ondrej-fabry, line 430: This should be named `GetRunningAgent` to be more descriptive about what it does. Function name `AlreadyRunningAgent` seems more like it would return `bool` value if the agent is running already.

tests/e2e/101_rest_api_test.go
```diff
@@ -27,7 +27,7 @@ func TestInfoVersionHandler(t *testing.T) {
 	ctx := Setup(t)
 	defer ctx.Teardown()
 
-	version, err := ctx.agentClient.AgentVersion(context.Background())
+	version, err := ctx.Agent.client.AgentVersion(context.Background())
```
ondrej-fabry, line 30: The client should be accessed via an exported method to make the planned separation easier.

ondrej-fabry (approved):

PR #1876, opened 2022-09-30
fix: Make agentctl get return config set by agentctl update

pemoticak:
After running `agentctl update` the updated config items will have `"io.ligato.from-client=agentctl"` (subject to change) label.
This behavior can be overwritten by using `--labels` flag on both `agentctl update` (update only labels specified by user ignoring the `"io.ligato.from-client=agentctl"`). When using `agentctl get` the label key can also be prefixed with `!` (`--labels="!negated=label"`) to exclude the config items with those labels from the result.


cmd/agentctl/commands/config.go
```diff
@@ -108,6 +111,9 @@ func runConfigGet(cli agentcli.Cli, opts ConfigGetOptions) error {
 	if err != nil {
 		return fmt.Errorf("parsing labels failed: %w", err)
 	}
+	if len(opts.Labels) == 0 && !opts.All {
+		labels["io.ligato.from-client"] = "agentctl"
```
ondrej-fabry, line 115: I do not think this is a good idea for a default filtering. For the update command it's definitely good to add that info. However in most scenarios agentctl works on top of other configuration and this behavior might confuse users if they see empty configuration when running `agentctl config get` first time.

pemoticak: Done.

ondrej-fabry (approved):

PR #1919, opened 2023-03-09
feat: Allow modification of default telemetry config, add metrics

samuel-dudik:
This PR adds the option to modify the default config of the telemetry plugin.

Additionally 2 new metrics were added - operational and admin state of available interfaces.

### DIFF
--- a/plugins/telemetry/config.go
+++ b/plugins/telemetry/config.go
@@ -36,17 +36,18 @@ type Config struct {
 	Skipped []string `json:"skipped"`
 }
 
-func defaultConfig() *Config {
+var DefaultConfig = func() *Config {
 	return &Config{
 		PollingInterval: defaultUpdatePeriod,
-		// TODO Even if disabled, the metric endpoint is still active. Consider shutting it down completely (discuss).
-		Disabled: true,
 	}
 }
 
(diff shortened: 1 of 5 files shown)


plugins/telemetry/config.go
```diff
@@ -36,10 +36,17 @@ type Config struct {
 	Skipped []string `json:"skipped"`
 }
 
+var ModifyDefaultConfig = func(config *Config) {
```
ondrej-fabry, line 39: How is this different from making `defaultConfig` exported variable?

plugins/telemetry/config.go
```diff
@@ -36,26 +36,26 @@ type Config struct {
...
+var DefaultConfig = func() *Config {
 	return &Config{
 		PollingInterval: defaultUpdatePeriod,
 	}
 }
 
 // loadConfig returns telemetry plugin file configuration if exists
 func (p *Plugin) loadConfig() (*Config, error) {
-	cfg := defaultConfig()
+	cfg := DefaultConfig()
```
ondrej-fabry, line 47: Just to be sure, add check if the `DefaultConfig` is nil. If it is, use zero value of `Config`.

PR #1921, opened 2023-03-22
Implementing support for nat44ed vrf table and route

Giluerre:
This patch provides support for nat44 vrf table and route vpp calls which were added to vpp v22.10.

However to succesfully use those calls, is needed to add a patch https://gerrit.fd.io/r/c/vpp/+/38459 
the addition of which is still pending.


plugins/vpp/natplugin/descriptor/nat44_vrf.go
```diff
@@ -0,0 +1,114 @@
...
+//ErrNATInterfaceFeatureCollision = errors.New("VPP NAT interface feature collision")
+
+// ErrDuplicateNATAddress is returned when VPP NAT address pool contains duplicate
+// IP addresses.
+// ErrDuplicateNATAddress = errors.New("Duplicate VPP NAT address")
+)
+
+// NAT44GlobalDescriptor teaches KVScheduler how to configure global options for
+// VPP NAT44.
+
```
fgschwan, line 46: I guess this is forgotten copy/paste... Just add godoc for Descriptor and it should be fine.

plugins/vpp/natplugin/descriptor/nat44_vrf.go
```diff
@@ -0,0 +1,114 @@
...
+type NAT44VrfDescriptor struct {
+	log        logging.Logger
+	natHandler vppcalls.NatVppAPI
+
+	// UseDeprecatedAPI tracks whether deprecated global API (NAT interfaces, addresses) is being used on NB.
+	// Used to orchestrate which data should be dumped from which descriptor on Retrieve.
+	UseDeprecatedAPI bool
```
fgschwan, line 53: The comment is not really up-to-date (another copy/paste). Do you need the bool variable itself?

plugins/vpp/natplugin/descriptor/nat44_vrf.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// NewNAT44GlobalDescriptor creates a new instance of the NAT44Global descriptor.
```
fgschwan, line 56: Please check your godoc/leftovers from copy/pasting in descriptor files. I will not point further ocurencies of that.

plugins/vpp/natplugin/descriptor/nat44_vrf.go
```diff
@@ -0,0 +1,114 @@
...
+	retrieved []adapter.NAT44VrfKVWithMetadata, err error) {
+
+	natTables, err := d.natHandler.Nat44VrfTablesDump()
+	if err != nil {
+		return nil, err
+	}
+	for _, natEntry := range natTables {
+		retrieved = append(retrieved, adapter.NAT44VrfKVWithMetadata{
+			//Key: nat.Nat44VrfTableKey(natEntry.TableVrfId),
+			Key:    strconv.Itoa(int(natEntry.TableVrfId)),
```
fgschwan, line 108: Why not use the key creator for given modeled configuration (the above commented out line)? This is duplicity and can lead to errors as model definition (the name tepmlate) is the source of truth how the key should be formatted.

plugins/vpp/natplugin/descriptor/nat44_vrf.go
```diff
@@ -0,0 +1,114 @@
...
+			Value:  natEntry,
+			Origin: kvs.FromNB,
```
fgschwan, line 110: I think that there should be something like fromSB as the data are clearly dumped from VPP. (SB=Southbound=VPP, NB=Northbound=userconfiguration)

plugins/vpp/natplugin/descriptor/nat44_vrf_route.go
```diff
@@ -0,0 +1,86 @@
...
+	typedDescr := &adapter.NAT44VrfRouteDescriptor{
+		Name:                 NAT44VrfRouteDescriptorName,
+		NBKeyPrefix:          nat.ModelNat44VrfRoute.KeyPrefix(),
+		ValueTypeName:        nat.ModelNat44VrfRoute.ProtoName(),
+		KeySelector:          nat.ModelNat44VrfRoute.IsKeyValid,
+		KeyLabel:             nat.ModelNat44VrfRoute.StripKeyPrefix,
+		Create:               ctx.Create,
+		Delete:               ctx.Delete,
+		Dependencies:         ctx.Dependencies,
+		RetrieveDependencies: []string{NAT44VrfDescriptorName},
```
fgschwan, line 53: Not sure about whether this one is needed as not retrieve is implemented in this descriptor.

fgschwan (changes requested):

proto/ligato/vpp/nat/nat.proto
```diff
@@ -178,3 +178,22 @@ message VirtualReassembly {
     // If set to true fragments are dropped, translated otherwise.
     bool drop_fragments = 4;
 }
+// Nat44EdVrfTable defines a vrf table for NAT44.
+message Nat44EdVrfTable {
```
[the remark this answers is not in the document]
Giluerre: It's working only with Endpoint dependent NAT44. That's the reason why I didn't generalize the naming to Nat44VrfTable.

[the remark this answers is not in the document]
ondrej-fabry: Thing is that we don't currently make distinction in our Proto API between in endpoint dependent vs. independent, beside the one boolean in global config. I think you could name the message `Nat44VrfTable` and provide better comment that includes info that it is used for ED NAT only. Additionally to this, you could add warning in log informing that the VRF table is only used if ED NAT is enabled.

Looking a bit out of the box, it might be that our Proto API might need a bit of refactor (which actually might already be in our backlog).. I am wondering if both ED and EI NAT can be enabled at the same time. In that case we definitely need some kind of refactor for the API to allow this.

proto/ligato/vpp/nat/nat.proto
```diff
@@ -178,3 +178,22 @@ message VirtualReassembly {
...
+
+}
+// Nat44EdVrfRoute defines a route for NAT44 VRF tables.
+message Nat44EdVrfRoute {
+    // Id of the source vrf
+    uint32 table_vrf_id = 1;
+    // Id of the destination vrf
+    uint32 vrf_id = 2;
+    // add_del
+    bool is_add = 3;
```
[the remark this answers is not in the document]
ondrej-fabry: @Giluerre Please, note that Proto API is not a copy of binary API. If it was, we would simply generate it. It is a more high-level API that should aim to be more generic and defined in a way that is as much future-proof as it can.

plugins/vpp/natplugin/descriptor/nat44_vrf_route.go
ondrej-fabry: However, it might be implemented in the future. If it does not prevent normal function let's keep it. It helps to make sure the VRFs are always retrieved before routes, since the route might be dependent on some VRF and having it in reverse order would fail to resolve the VRF.

proto/ligato/vpp/nat/nat.proto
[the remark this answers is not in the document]
pemoticak: Agreed, I think we should just call it `Nat44VrfTable` (+ explanatory comments) to keep it in line with current conventions. In the future refactoring of NAT seems likely, as it looks like to me that ED NAT and EI NAT features are diverging in VPP as time goes on. But without thorough investigation of these NAT modes in VPP (similarities/differences, use cases for each) it is hard to make the call on how should we approach the refactor - Split it in the Proto API? Or just creating 2 separate NAT VPP handlers would be enough? etc. Anyway, I am not knowledgeable enough in this area to have any strong opinions...

pemoticak:
Nothing else to add from my side, LGTM.

ondrej-fabry:
How did you re-generate Protobuf API? Also, it seems the `.proto` files have been reformatted? I assume it was your IDE that re-fromatted entire file.

plugins/vpp/natplugin/descriptor/nat44_vrf_route.go
```diff
@@ -0,0 +1,85 @@
...
+	err := d.natHandler.DelNat44VrfRoute(vrfRoute.SrcVrfId, vrfRoute.DestVrfId)
+	if err != nil {
+		d.log.Error(err)
+		return err
+	}
+	return nil
+}
+
+func (d *NAT44VrfRouteDescriptor) Dependencies(key string, natVrfTable *nat.Nat44VrfRoute) (deps []kvs.Dependency) {
+
```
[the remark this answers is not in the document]
Giluerre: I don't think that it's needed. SrcVrfId is table id of NatVrfTable and is only requirement to add NatVrfRoute. DestVrfId is id of normal vrf table.(In case the vrf table doesn't exist, the table is created, but this part is handled by vpp)

Giluerre:
> How did you re-generate Protobuf API? Also, it seems the `.proto` files have been reformatted? I assume it was your IDE that re-fromatted entire file.

make generate . You are right I didn't notice that .proto files were reformatted.

PR #1925, opened 2023-04-21
feat: Model rework

pemoticak:
This PR unifies remote registry and local registry into one structure containing all models directly. Remote and local models are also unified into one structure - `knownModel`. Handling of some properties (e.g. model name template) still depends on whether the model has a go type or not (it has a go type if the model was registered as a message from compiled `.pb.go` file, else if it was registered as a `dynamicpb.Message` it lacks a go type) but this could be eliminated in the future more easily then previously. It also adds proto extension that can be used to embed model specification and name template directly into `.proto` files. These changes make it possible to register models during vpp-agent startup directly from provided `.proto` files if desired. Ability of `agentctl` command to operate with models without go type is improved as well.


pkg/models/registry.go
```diff
@@ -38,210 +37,116 @@ var (
 // LocalRegistry defines model registry for managing registered local models. Local models are locally compiled into
 // the program binary and hence some additional information in compare to remote models, i.e. go type.
 type LocalRegistry struct {
```
fgschwan, line 39: LocalRegistry was named "local" to differentiate from RemoteRegistry, but that doesn't exist anymore. So it is little bit confusing why "local" (if i didn't know story behind it).  Renaming it to Registry clash with the interface Registry handling all previous instances of Local/RemoteRegistry. But i guess we can make it private by lower capital and its usage replace with Register interface.

proto/ligato/generic/options.proto
```diff
@@ -1,4 +1,4 @@
-syntax = "proto2";
+syntax = "proto3";
```
pemoticak, line 1: > CI proto check is failing (https://github.com/ligato/vpp-agent/actions/runs/4828274020/jobs/8601897603?pr=1925)

This is probably the culprit, because it is a breaking change. I think it was overlooked when migrating from proto v2 to proto v3, so I changed it now. Should not really affect anything, but I can revert this back to proto2 if needed.

fgschwan (changes requested):

proto/ligato/generic/options.proto
fgschwan: Ok, so the proto lint check is broken. I guess we can either fix it in this PR (if you see some easy solution) or what imho would be cleaner (in case of complicated/nonobvious solution) to revert back to proto2(and have lint capability) and make a issue that handles this exact problem in another PR.

pemoticak: Hmmm, I don't think the lint check is broken, it seems to be supposed to fail when it finds a breaking change in `.proto` files. So it found one and it failed. Probably to make sure breaking proto changes are not merged without discussion (maybe not merged at all)? @ondrej-fabry any thoughts on this?

fgschwan: @pemoticak You are right. I misread the log (https://github.com/ligato/vpp-agent/actions/runs/5012550482/jobs/8984628112?pr=1925). The `buf check lint` didn't fail. The `buf check breaking` running after that failed. So it is only check for making breaking changes in proto files. However, it is ok to change proto2 to proto3 in this case so we can ignore that check fail.

